PR #3, opened 2018-05-08
@angular and @angular/cli version 6.0.0 upgrade.

bthharper:
Upgraded the **enterprise-ng** components to the latest angular 6 release, see [Angular 6](https://blog.angular.io/version-6-of-angular-now-available-cc56b0efa7a4).

Any teams taking this component library would also need to upgrade their applications, although relatively straight forward for the control library it may be more complicated for complex applications.

The angular team have provided some tools to assist, which can be accessed from [Angular Update](http://update.angular.io/).

This PR currently still includes the rxjs compatibility package, which I will look at moving in a separate PR.


bthharper:
We should probably hold back on merging this until there is a consensus with dev teams.

src/app/datagrid/datagrid-paging-service-demo.service.ts
```diff
@@ -25,7 +25,7 @@ export class DataGridPagingServiceDemoService extends SohoDataGridService {
     const end: number = begin + req.pagesize;
     const data: Array<any> = this.data.slice(begin, end);
     const result: any = { total: 100, data: data };
-    return Observable.of(result);
+    return observableOf(result);
```
pwpatton: I see above you've used just `of` and here you're using `observableOf`. Why the difference?

src/app/tree/tree-demo.service.ts
```diff
@@ -1,7 +1,7 @@
+
+import {of as observableOf,  Observable } from 'rxjs';
```
pwpatton: ah, I think I found the answer to my question above.

bthharper: Yeah - they removed it - not sure why I like the other way.  I should really fix the others aliases too.

clepore:
I reverted `master` at @bthharper 's request so he can update his branch a bit and then he'll open a new PR for this 👍

PR #8, opened 2018-05-09
Adding travis ci

clepore:
Hey guys, check this out. We added the Travis CI stuff and code coverage graphs with drill downs.

Also note that I had to switch to using https://www.npmjs.com/package/ids-enterprise (our current master) to get this to work as Travis can't get behind our firewalls.


.travis.yml
```diff
@@ -12,7 +9,7 @@ install:
 before_script:
   - npm install -g angular-cli codecov
 script:
-  - npm build
+  - npm run build
```
tmcconechy: On the old build we decided to do both

npm run build --aot --prod
and 
npm run build

Basic there was errors with a prod build vs dev build (tree shaking and such). Might be a good idea to let that run here as well?

clepore: Testing both in a row like before though, would give you:
```
npm build  =>          // sohoxi.js = cat
npm build --prod =>    // sohoxi.js = dog
npm lint              // linting dog
npm test              // testing dog
```
...so you'd never actually test cat. If anything we'd have to do:
```
script:
  npm build
  npm run lint
  npm test
  npm build -aot --prod
  npm run lint
  npm test
```

Its up to you?

clepore: ℹ️ After a slack chat, @tmcconechy and I decided we should test the prod build to avoid any sort of tree-shaking oddities like the team saw before 👍

tmcconechy:
If both doesnt take that much longer maybe do both prod first. If it does then i guess we can skip it for now. It seems like it would catch some issues from what we saw but its not as critical i guess.

clepore:
It's okay @pwpatton we've all been there 🤓 

Let's do codecov for now with `npm run build` then we can see if we hit snags?

PR #20, opened 2018-05-16
Cleanup the root pkgjson

clepore:

- [x] This moves the actual `prepublishOnly` npm script cmd to the true package.json file used for publishing.
- [x] I also removed some deprecated command and renamed `npm run publish` so that we don't confuse that with the native `npm publish`


tmcconechy:
Are we going to need to use both gulp and grunt? Or are you planning to refactor ALL the grunt stuff to gulp?

We had planned on using npm scripts if we are refactoring to anything...

PR #21, opened 2018-05-17
feature/SOHO-7935-add-findRowsByValue-to-datagrid-control

micheal-cao:
expose findRowsByValue in angular datagrid component


bthharper:
Looks okay to me.

PR #22, opened 2018-05-17
Soho 6787 Removed NgModel from providers of SohoDropdownComponent

bthharper:
Removed the `NgModel` from the providers list, and replaced with an adapted `ControlValueAccessor` set on the `NgControl` (if present).  This allows the `SohoDropdownComponent` to notify any associated model of changes, and also construct the correct option "value" - which is generated internally by the standard `Select[Single|Multiple]ControlValueAccessor` directives.

Issues with clients of `dropdown` using `source` as a means of obtaining their data should also work, wit the addition of an event handler for `requestend` which pushes an `updated` request for the dropdown in the execution pipeline.

Code using `source` with `SohoDropdownComponent` should pass 'true' into the `response` callback to avoid the jQuery control creating it's own `option` elements.  These should instead be handled by Angular.

```typescript
    const states = [
      { value: 'AK', label: 'Alaska'},
      { value: 'AZ', label: 'Arizona'},
      { value: 'CA', label: 'California'},
      { value: 'CO', label: 'Colorado'},
      { value: 'MN', label: 'Minnesota'},
      { value: 'ND', label: 'North Dakota'},
      { value: 'OR', label: 'Oregon'},
      { value: 'WA', label: 'Washington'},
      { value: 'WY', label: 'Wyoming'}
    ];

    this.sourceoptions = states;
    response(states, true);
```

I would not recommend using the `source` option, but would favour the use of Inputs.

Other things worth noting:

* removed the Landmark example, as this uses a hybrid of reactive and template driven forms which is not recommended, and Landmark no longer does them this way.
* added a new Reactive Forms demo


src/soho/dropdown/soho-dropdown.component.ts
```diff
@@ -1,33 +1,49 @@
...
+  NgModel,
+  NgControl,
+  ControlValueAccessor
+} from '@angular/forms';
 
 @Component({
   selector: 'select[soho-dropdown]', // tslint:disable-line
-  template: '<ng-content></ng-content>',
-  providers: [NgModel],
-  changeDetection: ChangeDetectionStrategy.OnPush
```
fitzorama: Curious to why this is being removed; I think it is something we want to keep.

Additionally, we still need the change for @Input set source() where the source function is wrapped to run inside of Angular, is that correct?

bthharper: The JIRA was to remove the `NgModel` from the `providers` list, as this was causing problems for another team and their build.  The idea was to replace this with a custom `ControlValueAccessor` which would provide the glue between the Soho (IDS) control and Angular.  This proved complicated, as the value accessors needed to be specific for each selection type so I decided to wrap the existing ones allowing me to grab the `changeFn` to invoke from within the `changed` event.

I will add back `changeDetection: ChangeDetectionStrategy.OnPush`, I must have removed it when removing `ChangeDetectorRef` from the constructor (which I no longer used).

The problem with forcing the `source` method to run inside Angular, is that this causes all the code invoked as part of the source/response interaction also to run inside the Angular zone.  I found this caused spurious changes in any associated Form Model, and broke multi-selection.

I found that by adding a handler for the `requestEnd` I was able to force an update, such that the dropdown was updated correctly, this seemed to negate the need for the `source` delegate.

This also removed the need for the caller to invoke `markForCheck`.

I would really appreciate you giving this a go in your code, as different usages introduce different issues.

bthharper:
Last push fixes the issue of dropping the selected item when calling source.

bthharper:
@fitzorama @clepore @pwpatton @tmcconechy this is ready now.

PR #29, opened 2018-05-23
SOHO-8005 - Add 'tag' support to the IDS Angular components

bthharper:
Added a Straw Man implementation of tags for comment.


src/soho/tag/soho-tag.component.ts
```diff
@@ -0,0 +1,200 @@
...
+        .on('click', (e: JQuery.Event) => this.onClick(e) );
+    });
+  }
+
+  private onBeforeTagRemove(e: JQuery.Event, element: HTMLElement) {
+    this.ngZone.run(() => this.beforeRemove.next(e) );
+  }
+
+  private onClick(e: JQuery.Event) {
+    this.ngZone.run(() => this.click.next(e) );
```
pwpatton: Depending on when these are executed they may not cause change detection.
For example in soho-tabs we have the events hooked up before the initialization of the tabs control so that we get the initial "tabs activated event". 

In that case we needed to add a setTimeout to the ngZone.run() so change detection would fire.
```
this.ngZone.run(() => setTimeout(() => this.tabActivated.next(e), 1));
```

If that's not the case then this should be fine as user initiated events will initiate change detection. The only other thing that could be an issue is if events are fired from soho w/o any user event being the cause.

src/soho/tag/soho-tag.component.ts
```diff
@@ -0,0 +1,200 @@
...
+  Output,
+  EventEmitter
+} from '@angular/core';
+
+export type SohoTagType = 'error' | 'good' | 'alert' | 'secondary' | undefined;
+
+@Component({
+  selector: '[soho-tag-list]', // tslint:disable-line
+  template: '<ng-content></ng-content>'
+})
```
pwpatton: Should be OnPush
```
changeDetection: ChangeDetectionStrategy.OnPush
```

src/soho/tag/soho-tag.d.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+/**
+ * Rating Options
+ */
+interface SohoTagOptions {
+}
+
+/**
+ * Rating Api.
```
fitzorama: not Rating Api

src/soho/tag/soho-tag.d.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * Infor Design Enterprise (fna sohoxi) 'Tags'.
+ *
+ * This file contains the Typescript mappings for the public
+ * interface of the jQuery Tags control.
+ */
+
+/**
+ * Rating Options
```
fitzorama: SohoTag Options

src/soho/tag/soho-tag.d.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Rating Api.
+ */
+interface SohoTag {
+  /** The settings option */
+  settings: SohoTagOptions;
+
+  /** Updates the rating with any new settings and data */
+  updated(settings?: SohoTagOptions): void;
```
fitzorama: Updates the tag

bthharper: Not that I copied rating!

fitzorama: I believe you :-)

PR #30, opened 2018-05-24
Updating ids-enterprise package to the 20180523 nightly build.

pwpatton:


tmcconechy:
Once this is merged i'll pull this onto my PR in the other branch and it should fix that.

tmcconechy:
OK, maybe caching problem i see it now :grin:

Issue #32, opened 2018-05-29
Busy Indicator stuck on loading


tmcconechy:
Issue moved from https://jira.infor.com/browse/SOHO-8033

Toggling the "activated" state of the busy indicator quickly can sometimes cause it to be stuck in one state. Possibly only during component initiation.

Code:

```
// Component template
<div soho-busyindicator [activated]="isBusy" [blockUI]="true" [displayDelay]="0">

// Component controller
class MyComponent {
  isBusy: boolean;

  ngOnInit() {
    this.isBusy = true;
    fetchSomething().then(() => this.isBusy = false);   // Busy indicator remains "activated" after this
  }
}

```

SOHO version:

```
$ cat package.json | grep ids-enterprise
    "ids-enterprise": "4.7.0-beta",
    "ids-enterprise-ng": "4.7.0-dev.20180523",

```

The problem appears to be the asynchronous handling of `initiallyActive` in `ngAfterViewInit`` in soho-busyindicator.directive.ts. Basically, if the `activated` input is changed between the callbacks for `this.ngZone.run(() => setTimeout(() => ...))`, then it will still get stuck with the "initiallyActive" state.

I tried fixing it locally (http://git.infor.com/projects/ITECH/repos/homepages/commits/d30cc35d6da4a54df6b90a11d9de5c6e4e5b5846), and it seems to solve my problem. But I don't know if it would cause new problems elsewhere.. For instance -- Why is there a setTimeout there to begin with?

tralarn:
@tmcconechy , would be really good to get this fixed for 4.7.0. Related to Homepages 12.0.25 release.

tmcconechy:
Based on the comments on http://git.infor.com/projects/ITECH/repos/homepages/commits/d30cc35d6da4a54df6b90a11d9de5c6e4e5b5846

It seems we dont need the timeout. If you want to push a pull request here , we can go with your fix.

@clepore we didnt make a 4.7.x branch so if he pr's onto master will this be included in 4.7.0 this week?

clepore:
That's up to the devs in the NG team and if it's ready.

tmcconechy:
Ok, Cool. So not sure of andreas github details but if he can get a PR in with this change we should be good for 4.7 to review and add.

bthharper:
[...]

PR #33, opened 2018-05-29
Fix initial state of busyindicator

anhallbe:
My first PR in this repo. Please review the code, test and see if there's anything I've missed with labels/issue-links/naming etc.

This should make sure the "activated" state is correct when the
component is initiated. Fixes LIME-3617, SOHO-8033 and #32

### DIFF
--- a/src/soho/busyindicator/soho-busyindicator.directive.ts
+++ b/src/soho/busyindicator/soho-busyindicator.directive.ts
@@ -216,7 +216,7 @@ export class SohoBusyIndicatorDirective implements AfterViewInit, AfterViewCheck
       // Initialise any event handlers.
       this.hookupRegisteredEvents();
 
-      // Initial busy state if set to true
+      // Set initial state
       // get back into the angular zone so the setTimeout will trigger change detection.
       this.ngZone.run(() => this.activated = this.initiallyActive);
     });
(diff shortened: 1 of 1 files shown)


src/soho/busyindicator/soho-busyindicator.directive.ts
```diff
@@ -105,14 +105,11 @@ export class SohoBusyIndicatorDirective implements AfterViewInit, AfterViewCheck
   /** Controls the activated state of the busy indicator. */
   @Input()
   public set activated(value: boolean) {
-    if (this.busyindicator) {
-      if (value) {
-        this.open();
-      } else {
-        this.close(true);
-      }
+    this.initiallyActive = value;
```
bthharper: Do we need this here?  I assumed this was just for the `ngAfterViewInit` method?

anhallbe: Yes, I think so

Example scenario:

1. set activated true
2. ngAfterViewInit called
3. this.busyindicator is set
4. ngZone.run is called (with async callback)
5. set activated false (initiallyActive is still true, since this.busyindicator is truthy)
6. ngZone.run callback is called, which sets activated to initiallyActive, which is true)<
7. State is now "activated", even though it was set to false.

1 and 5 is performed by parent. Does that make sense?

PR #79, opened 2018-05-31
Added some missing color picker settings

tmcconechy:
Added some missing color picker options noted on https://jira.infor.com/browse/SOHO-8007


src/soho/colorpicker/soho-colorpicker.component.ts
```diff
@@ -88,6 +96,70 @@ export class SohoColorPickerComponent extends BaseControlValueAccessor<any> impl
...
+  @Input() set uppercase(value: boolean) {
+    this.isUppercase = value;
+    this.options.uppercase = value;
+
+    if (this.colorpicker) {
+      this.colorpicker.settings.uppercase = value;
+    }
+
+    if (this.colorpicker) {
+      this.colorpicker.updated(this.options);
```
pwpatton: would like to see a `callUpdatedAfterCheck: boolean` flag then call the updated() function in a runOutsideAngular in ngAfterViewChecked.

Example
In the @Inputs just set the flag.
```
  if (this.colorpicker) {
      this.callUpdatedAfterCheck = true;
  }
```

In ngAfterViewChecked check the flag and then call updated and reset the flag
```
ngAfterViewChecked() {
  if (this.callUpdatedAfterCheck) {
    this.ngZone.runOutsideAngular(() => this.colorpicker.updated(this.options));
    this.callUpdatedAfterCheck = false;
  }
}
```

Maybe not super important right now but it's a good way to ensure that updated() is not called for each input setting but instead at the end of a tick() after all inputs have been set.

tmcconechy: I can make the change. Is there a reference component done with all ngZone updates and the change to only trigger on ngAfterViewChecked ?

bthharper: I would take a look at dropdown, as this has both the run updated on check boolean, and the NgZone code.

bthharper:
Looking at the `SohoColorPickerComponentValueAccessorDelegator `, which for obvious reasons mimics the dropdown implementation, I wonder if there is a case for creating a generic delegate used by both.  This looks like a repeating pattern.

src/soho/colorpicker/soho-colorpicker.component.ts
```diff
@@ -210,4 +342,74 @@ export class SohoColorPickerComponent extends BaseControlValueAccessor<any> impl
...
+/**
+ * Provides a 'wrapper' around the {ControlValueAccessor} added by
+ * angular when handling `select` elements.
+ *
+ * This class allows the {SohoColorPickerComponent} to interoperate with
+ * the {ControlValueAccessor}.  Specifically, providing access to the
+ * onChange function, which we must call when the value of the colorpicker
+ * is modified.
+ *
+ * It also exposes the encoding used storing complex objects as
```
bthharper: Delete this line (and the next)?

pwpatton: not sure, why these lines?

src/soho/colorpicker/soho-colorpicker.component.ts
```diff
@@ -79,15 +104,67 @@ export class SohoColorPickerComponent extends BaseControlValueAccessor<any> impl
...
       } else {
-        this.colorpicker.readonly();
+        this.ngZone.runOutsideAngular(() => this.colorpicker.readonly());
         this.isEditable = false;
       }
     }
   }
 
+  /**
+   * Enables or disables forceing upper case hex.
```
fitzorama: forcing

tmcconechy:
Fixed spelling please re-approve.

PR #87, opened 2018-06-07
Updated types for hierarchy event to match the hierarchy control API

krishollenbeck:

Updates types for hierarchy event to match new properties added to the hierarchy control API.

https://github.com/infor-design/enterprise/pull/123/commits/693fef8d3ead5d8c314ecca6e9204f21337d1ab0

> Other Details:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "dev": true,
       "requires": {
         "@angular-devkit/core": "0.6.0",
-        "rxjs": "^6.0.0"
+        "rxjs": "6.2.0"
       }
     },
     "@angular-devkit/build-angular": {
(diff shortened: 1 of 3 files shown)


krishollenbeck:
@tmcconechy, I wasn't too sure about those. How do we want to handle those going forward? I was under the impression that they should always be promoted. https://yarnpkg.com/lang/en/docs/yarn-lock/#toc-check-into-source-control

tmcconechy:
I feel like they should only be promoted if a dependency changed or if you are on the latest npm (updated to 6.10 fx). Which version of npm are you on? The issue is one version added the ^ so if we are on different versions we just end up changing it back and forth every time?

krishollenbeck:
Okay, I reverted them.

Issue #88, opened 2018-06-07
Font weight for step process control is inconsistent.


krishollenbeck:
**Describe the bug**

**To Reproduce**
1. Launch enterprise-ng app and go to localhost:4203/step-process
2. Resize screen to a mobile view so the footer buttons appear at the bottom of the page.
3. Notice the font weights are not the same.

**Expected behavior**

The font weights should be the same.

**Screenshots**

https://imgur.com/a/2Ixzk9n

**Platform**
 - Device (if applicable) [e.g. iPhone 6 or Samsung Galaxy S6]
 - OS Version: [e.g. Windows 10 or iOS 8]
 - Browser Name [e.g. chrome, safari, stock browser]
 - Browser Version [e.g. 22, 66.0.3359.181 (Official Build) (64-bit)]

**Additional context**

PR #89, opened 2018-06-07
Fix for step control footer buttons font weight issue.

krishollenbeck:

Fix for step control footer buttons font weight issue.

https://github.com/infor-design/enterprise-ng/issues/88

1. Follow steps in related issue link.
2. Apply change to see it fixed.

> Other Details:

The underlying issue here is that the `soho-button` directive adds the class ` btn-secondary` after the class `js-btn-save-changes`. And in order for the button to style properly the `btn-secondary` class needs to be the first class on the element.

### DIFF
--- a/src/soho/stepprocess/soho-stepprocess.component.html
+++ b/src/soho/stepprocess/soho-stepprocess.component.html
@@ -26,7 +26,7 @@
   <!-- STEP CONTROL PHONE CONTROLS -->
   <ng-content select="div[soho-step-content]"></ng-content>
     <div class="phone-action-bar phone-visible">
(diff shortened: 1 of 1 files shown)


bthharper:
I thought starting a review stopped merging?

tmcconechy:
Did he merge before you reviewed?

bthharper:
Can't see a time on my review, but it's before your reviews so I assume so.

bthharper:
Doesn't really matter, but was interested if we could do this more cleanly.

tmcconechy:
OK, you can still comment and we can still discuss. Just click the code lines on the side of the code and do the single comments.

Issue #42, opened 2018-05-29
SOHO-7944: sohoxi-angular library should not include .spec.ts files


clepore:
I recognized that in sohoxi-angular, there are some test files (.spec.ts) included in the package. This will increase the bundle size and the download size of the whole package. Test files should be excluded from npm publish.

*Expectation*:
there should be no test files in sohoxi-angular package.

AhnpGit:
so now I think instead of creating JIRA on SOHO board, I'll just create issues here ? 👍

tmcconechy:
We are moving to migrate all issues to github but started with just the angular ones so we can get more contributions and eyes on them. So for now

- Angular issue -> can create here
- Core code issue -> still user Jira
- Not sure which -> still user Jira

We will migrate core code (https://github.com/infor-design/enterprise) eventually but there are too many issues in the backlog now so we need to take care of some of them before we make the switch.

bthharper:
@clepore  We close this now?

PR #92, opened 2018-06-08
Use 1 package.json file in root for simplicity

clepore:
**Explain the details for making this change. What existing problem does the pull request solve?**

Combine the package.json files so we do not have to worry about manually updating packages, etc.

Compare the output of `npm pack publish/` on master:
old-master

with the output of `npm pack` on this branch with 1 package.json
new-one-pkgjson

This is not a **must** PR. If you don't like the change in file structure, comment or close this, no hard feelings 👍


tmcconechy:
Note that this closes #42

package.json
```diff
@@ -1,18 +1,40 @@
...
   "scripts": {
     "ng": "ng",
     "start": "ng serve",
     "build": "ng build",
     "test": "ng test --browsers=PhantomJS --code-coverage",
     "lint": "ng lint",
     "e2e": "ng e2e",
-    "publish-setup": "cd ./publish && npm run prepublishOnly && cd ../",
+    "prepublishOnly": "npx gulp publish",
     "publish-dev": "npm publish publish/ --tag dev",
```
tmcconechy: We don’t need these now?

clepore: Remember the green means something was added. The red means something was removed. 

I removed the cmd (publish-setup) that calls a cmd (prepublishOnly) from the now nonexistent publish/package.json, and we now have that cmd (prepublishOnly) available directly.

Issue #86, opened 2018-06-06
Soho Input cannot be disabled using FormGroup


tomjohn9:
Steps to reproduce:
* Have Component with template with input `<input formControlName="count" soho-input>`
* Use following body of component .ts  
```
formGroup: FormGroup;
model: {
    count: number
  } = {
    count: 10
  };

  constructor(private formBuilder: FormBuilder) {
    this.formGroup = this.createForm();
    this.formGroup.disable();
  }

  private createForm() {
    return this.formBuilder.group({
      count: [this.model.count]
    });
  }
```

**Expected result**:
  input component is disabled

**Actual result:**
  input component is enabled

**Guidelines:**
got from channel chat - can help
The hook between an angular component and the jQuery widget does not handle disable requests.  Specifically, the components needs to override the following ControlValueAccessor method:

```
/**
 * This function is called when the control status changes to or from "DISABLED".
 * Depending on the value, it will enable or disable the appropriate DOM element.
 *
 * @param isDisabled
 */
 setDisabledState(isDisabled: boolean): void {
   // NOP
 }
```

bthharper:
I have a fix for this on my fork, however all other controls that implement the ControlValueAccessor need to be updated so will take a little longer.

tmcconechy:
This seems ok to me if noted in the changelog.

bthharper:
@tmcconechy do we assign these for QA?

tmcconechy:
We dont have a way for functional QA to QA any of the angular stuff at the moment. QA is crowd sourced among our teams for this part. But working on that. So just close the issue for now.

PR #94, opened 2018-06-12
Soho input cannot be disabled using form group

bthharper:

When using angular **ReactiveForms**, the **enterprise-ng** components do not respect the `disable` state on the form/group/control model or the initial value of any model specified in the model.

This is because the hook between an angular component and the jQuery widget does not handle disable requests.   Specifically, the components needs to override the following `ControlValueAccessor` method ` setDisabledState(isDisabled: boolean)`.

https://github.com/infor-design/enterprise-ng/issues/86

Added several UTs to ensure the forms models values are used by the components on initialization and when they are changed by angular.

Add new **Reactive Forms** examples, which includes the set of known components linked to a form model, and disabled by default.

> Other Details:

https://github.com/infor-design/enterprise-ng/issues/86


CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 ## Whats New
+* 2018-06-08 - BTH - Fixed component support for ReactiveForms, specifically for 'disabled' and 'value'.
 * 2018-06-01 - TJM - Updated colorpicker to use ngZone and added a few missing settings. This effects the constructor so may effect those using AOT.
```
tmcconechy: Can you add a note like this for any of the components you refactored to ngZone. I.E. touched the constructor.
As it requires an AOT recompile so were asked to explicitly note those.

package.json
```diff
@@ -12,29 +12,29 @@
...
+    "@angular/forms": "6.0.4",
+    "@angular/http": "6.0.4",
+    "@angular/platform-browser": "6.0.4",
+    "@angular/platform-browser-dynamic": "6.0.4",
+    "@angular/router": "6.0.4",
     "@types/lscache": "^1.0.7",
     "classlist.js": "^1.1.20150312",
     "core-js": "^2.4.1",
-    "ids-enterprise": "^4.8.0-dev.20180528",
+    "ids-enterprise": "^4.7.0-rc.2",
```
tmcconechy: https://www.npmjs.com/package/ids-enterprise/v/4.8.0-dev.20180611 is newer... Or 4.7.0 (not RC)

CHANGELOG.md
bthharper: Just merging these changes in now - my branch was taken before this.

package.json
bthharper: ... as above.  ;-)

CHANGELOG.md
bthharper: Not sure I understand the comment about requires an AOT recompile, do you have any information on that?  Is this Ming.le / homepages trying to drop angular components in?

tmcconechy: yes its mingle / home pages dropping in. my understanding is that if we change any component constructors they would like to know as they need to do something on their end @pwpatton may have a better understanding. I just think of it as making a note if we add something to a contructor / as we make the ngone changes so they can see that in the release notes.

PR #116, opened 2018-06-27
Add @types/jquery to dependencies

### LINKED ISSUE #115: Find a way to decouple the dependency on @types/jquery
**Describe the bug**
When setting up a project from scratch using `Angular CLI`  and `ids-enterprise-ng`, it is tricky to find the right version of `@types/jquery`. If installing the wrong version (e.g 3.3.x), `ids-enterprise-ng` won't work. What I observe is the correct version is **3.2.16**

**To Reproduce**
- Generate a project using `Angular CLI` v6.
- Install `ids-enterprise-ng`
- Install `@types/jquery@latest`
- Add this piece of code into your `tsconfig.app.json`
```
"include": [
    "**/*",
    "../node_modules/ids-enterprise-ng/**/*"
  ]
```
- Run `ng serve` and an error occurs related to `@types/jquery` doesn't match with Node

**Expected behavior**
We should add the correct version of `@types/jquery` to publish `package.json` to reduce this frustration and make it easier to setup a project.

**Screenshots**
image

**Platform**
 - Device: laptop
 - OS Version: Windows 10
 - Browser Name: Chrome
 - Browser Version: Version 67.0.3396.99 (Official Build) (64-bit)

**Additional context**
**N.A.**

ahnpnl:
This is just a temporary workaround for this issue. It helps the developers to setup easier from scratch.
This also automatically installs `@types/jquery` whenever the users install `ids-enterprise-ng` so they don't have to worry about missing dependencies, increasing the convenience.

Close #115

Follow the steps describing in #115

### DIFF
--- a/publish/package.json
+++ b/publish/package.json
@@ -25,9 +25,9 @@
   },
   "dependencies": {
     "@angular/cli": "6.0.7",
+    "@types/jquery": "3.3.1",
     "classlist.js": "^1.1.20150312",
     "ids-enterprise": "^4.8.0-dev.20180621",
-    "jquery": "3.3.1",
-    "@types/jquery": "3.3.1"
+    "jquery": "3.3.1"
   }
 }
--- a/publish/package.json
+++ b/publish/package.json
@@ -27,6 +27,7 @@
     "@angular/cli": "6.0.7",
     "classlist.js": "^1.1.20150312",
     "ids-enterprise": "^4.8.0-dev.20180621",
-    "jquery": "3.3.1"
+    "jquery": "3.3.1",
+    "@types/jquery": "~3.2.16"
   }
 }
--- a/publish/package.json
+++ b/publish/package.json
@@ -28,6 +28,6 @@
     "classlist.js": "^1.1.20150312",
     "ids-enterprise": "^4.8.0-dev.20180621",
     "jquery": "3.3.1",
-    "@types/jquery": "~3.2.16"
+    "@types/jquery": "3.3.1"
   }
 }


publish/package.json
```diff
@@ -27,6 +27,7 @@
     "@angular/cli": "6.0.7",
     "classlist.js": "^1.1.20150312",
     "ids-enterprise": "^4.8.0-dev.20180621",
-    "jquery": "3.3.1"
+    "jquery": "3.3.1",
+    "@types/jquery": "~3.2.16"
```
tmcconechy: Should we use the same version here? https://www.npmjs.com/package/@types/jquery/v/3.3.1

ahnpnl: I think yes it makes sense to keep the typing package version to be the same as the `jquery` version. Will change

publish/package.json
```diff
@@ -27,6 +27,7 @@
     "@angular/cli": "6.0.7",
     "classlist.js": "^1.1.20150312",
     "ids-enterprise": "^4.8.0-dev.20180621",
-    "jquery": "3.3.1"
+    "jquery": "3.3.1",
+    "@types/jquery": "3.3.1"
```
pwpatton: Can you move `@types/jquery` up to where the other types are included?
They are toward the top of the devDependencies

ahnpnl: do you mean it should be under @angular/cli to make it in alphabetical order ?

ahnpnl:
No, the purpose of this PR is when installing `ids-enterprise-ng`, it will automatically install the correct version of `@types/jquery`. It is inconvenience for the users when they don't know exactly which version of `@types/jquery` will work with `ids-enterprise-ng`. 

The situation is when installing `@types/jquery` at 3.3.4, the app can't build because it is not the expected version. After testing a while to find out that, I find it inconvenience. Therefore, we should help the users to setup easier by automatically install `@types/jquery` when installing `ids-enterprise-ng`.

If in the future, there is a way to solve this then we can take out `@types/jquery` as delivered dependencies.

pwpatton:
I see it's because we deliver source for the ng wrappers code and not compiled build npm package. Once we package it and deliver compiled then the dependency should get hooked up automatically w/o your devs having to know about it.

Issue #93, opened 2018-06-11
SOHO-7508 - Listview is not able to render items wrapped inside the ng-container


tmcconechy:
**Describe the bug**
When using ng-container inside soho listvew, list items are not displayed. Generated ul is empty. ng-container is used to be able to use ngFor and ngIf directive together.

**To Reproduce**

```
<soho-listview #multipleSelectListView selectable="single">
 <ng-container *ngFor="let item of items">
   <li *ngIf="item.test.length > 3" soho-listview-item>
      <p soho-listview-header>{{item.test}}</p>
    </li>
  </ng-container>
</soho-listview>
```

**Expected behavior**
So far I can go with different object only with visible items. But according to https://angular.io/guide/structural-directives#ng-container-to-the-rescue

The Angular <ng-container> is a grouping element that doesn't interfere with styles or layout because Angular doesn't put it in the DOM.

And as ng-container is standard approach in angular I think it's worth to investigate.

From: https://jira.infor.com/browse/SOHO-7508

PR #117, opened 2018-06-28
Listen for changes when the underlying list items change.

bthharper:

When using a component template to generate the list items in a list view, the selection check boxes are not being added.

Fixes https://github.com/infor-design/enterprise-ng/issues/93.

See Issue for example.


src/soho/dropdown/soho-dropdown.component.ts
```diff
@@ -387,7 +387,7 @@ export class SohoDropDownComponent implements AfterViewInit, AfterViewChecked, O
         // has been updated (assuming there is one), so
         // execute updated after angular has generated
         // the model and the view markup.
-        setTimeout(() => this.updated() );
+        setTimeout(() => this.updated());
```
pwpatton: Timeout shouldn’t be required here. But it shouldn’t hurt either since it’s outside Ng zone.

Issue #48, opened 2018-05-29
SOHO-7766: Add More Soho Types.


clepore:
Carying on from http://git.infor.com/projects/SOHO/repos/angular-components/pull-requests/623/overview

Some types should be add to angular like

- [ ] Soho.string.stripHTML (and any white list security functions)
- [ ] Personalize Colors / Set theme
- [ ] resetForm (and any missing Validation functions)
- [X] Soho.Locale - functions like parseDate ect

tmcconechy:
Part of this done on https://github.com/infor-design/enterprise-ng/pull/124#issuecomment-402163055

tmcconechy:
Descoping this issue for now.

PR #124, opened 2018-07-03
Fixed locale typings.

### LINKED ISSUE #122: Locale typings incorrect
**Describe the bug**

The `soho-locale.d.ts` file contains a declaration for `Locale` which should be inside the Soho global. 
 The correct `Locale` has been added in the declaration in `soho-datagrid.d.ts` which breaks encapsulation rules.

As Visual Studio uses the typings to provide intellisense, it incorrectly allows Locale to used used, which causes a runtime error.

There are also some missing methods/overloads in `Locale` typing itself, e.g. `translate(...)` should take two arguments.

**To Reproduce**

The following looks okay in the IDE (Visual Studio and Visual Studio Code) and compiles, but fails ar runtime.
```
Locale.set.('en-GB');
```

**Expected behavior**

```
Soho.Locale.set('en-GB');
```
**Screenshots**

N/A

**Platform**

All

**Additional context**

bthharper:

When adding internationalisation support to an application, the typings for the widget library define a global `Locale` object which is not defined at runtime.  Removal of this old declaration will realign the typings to the widget library, and reduce coding errors.

I have also taken the opportunity to move the `Soho` global into `soho.d.ts` making it more extensible.

Closes #122.

The following, compiles, but breaks at runtime:
```
Locale.translate('OK')
```

### DIFF
--- a/src/soho/datagrid/soho-datagrid.d.ts
+++ b/src/soho/datagrid/soho-datagrid.d.ts
@@ -358,19 +358,7 @@ type SohoDataGridColumnEditorFunction = (
   item?: any
 ) => SohoDataGridCellEditor;
 
-declare var Editors: {
-  // Supports, Text, Numeric, Integer via mask
-  Input: SohoDataGridColumnEditorFunction;
-  Textarea: SohoDataGridColumnEditorFunction;
-  Checkbox: SohoDataGridColumnEditorFunction;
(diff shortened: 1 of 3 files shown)


bthharper:
I think we can remove the `Locale` work from #48 - as most methods look to be exposed already, I just found the one method with a changed signature.

PR #154, opened 2018-08-03
Added script to test for node 10.

tmcconechy:

Since the package lock files are getting out of sync because we run different versions of node.
I put a check to insure that we all use a similar version.

Also updated the lock files.

Closes #540

- npm install (will fail if not on node 10)
- install node 10

MAC:
I suggest using https://github.com/creationix/nvm to handle the versions.
As its possible that you would need different versions for different projects

Windows:
You can use a similar version manager for windows https://github.com/coreybutler/nvm-windows
Or use the node installer https://nodejs.org/en

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "dev": true,
       "requires": {
         "@angular-devkit/core": "0.6.0",
-        "rxjs": "6.2.0"
+        "rxjs": "^6.0.0"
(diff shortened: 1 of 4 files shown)


tmcconechy:
Decided to not enforce this for Angular (yet?) based on feedback and discussion.
Will still enforce for enterprise tho.

PR #178, opened 2018-09-04
Datagrid events

pwpatton:
Fix for not getting the correct information from a soho-datagrid event.


src/app/datagrid/datagrid-dynamic.demo.ts
```diff
@@ -115,7 +115,7 @@ export class DataGridDynamicDemoComponent implements AfterViewInit {
     // select right click row, unselect any other row first.
     // -----------------------------------------------------------------
     this.dataGrid.unSelectAllRows();
-    this.dataGrid.selectRows(e.row);
+    this.dataGrid.selectRows([ e.row ]);
```
bthharper: This explains the issue with selectRows I was having.

PR #191, opened 2018-09-21
fix test failure

pwpatton:
For some reason a test started failing. The particular `it` test alone worked fine but some order of tests seems to be causing it to fail. Changing the order caused it to pass.

nothing really, just moving the test to a different order made it work. There is some deeper issue that need to be looked into maybe.

### DIFF
--- a/projects/ids-enterprise-ng/src/lib/datagrid/soho-datagrid.spec.ts
+++ b/projects/ids-enterprise-ng/src/lib/datagrid/soho-datagrid.spec.ts
@@ -298,6 +298,29 @@ describe('Soho DataGrid Render', () => {
     expect(component.datagrid.dataset).toBe(testData);
   });
 
+  it('fires `rowRemove` when removeSelected called.', (done) => {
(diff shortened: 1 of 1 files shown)


clepore:
If some tests are relying on one another to work before running the next test (i.e. in a series), we may need to add some more "reset" code to the `beforeEach()` and `afterEach()` functions to reset each "scenario" before the next test runs.

pwpatton:
I believe that's what resetting the test bed in angular is supposed to do.

https://github.com/infor-design/enterprise-ng/blob/b22ea80916a781a72cff834d8a5c72b7830186a7/projects/ids-enterprise-ng/src/lib/datagrid/soho-datagrid.spec.ts#L271-L284

clepore:
That's what I mean, it might not be resetting something we need for the one test, which is why it works when it comes first. It's just a stab it the dark.

bthharper:
I thought I'd fixed an issue where a static was being modified, I'll take a look.

PR #192, opened 2018-09-24
Datagrid unit test fix

bthharper:

Removed usage of static member which potentially caused cross unit test issues.

N/A

```ng test```


bthharper:
Was trying to understand the issue here, so have made a couple of changes and push the offending test back to its original location.

If someone could maybe try this out locally, as it works for me.

Issue #185, opened 2018-09-19
Memory leak in SohoContextMenuDirective


prkarlsson:
**Describe the bug**
The ngOnDestroy function in the SohoContextMenuDirective does not call the destroy function on the jQuery context menu. This causes leaks of detached DOM elements when creating and destroying Angular components that uses context menus. Note that there is a comment in the ngOnDestroy function that can also be found in SohoMenuButtonComponent so that component may be affected as well.

The leaked DOM elements can be found by using the Chrome Memory profiler. They show up as "Detached HTMLDivElement" etc.

**To Reproduce**
1. Create an Angular component that uses SohoContextMenuDirective 
2. Create and destroy the component multiple times in a single-page app
3. Take a heap snapshot in the Chrome Memory profiler
4. There are several detached DOM elements kept alive since the jQuery context menu is not destroyed.

**Expected behavior**
The ngOnDestroy function in SohoContextMenuDirective should call the destroy function for the jQuery context menu. There should be no detached DOM elements after destroy is called.

**Screenshots**
-

**Platform**
 - Device: N/A
 - OS Version: Windows
 - Browser Name: Chrome
 - Browser Version:  69.0.3497.92 (Official Build) (64-bit)

**Additional context**
The issue was found when using the enterprise-ng components in Infor Homepages.

pwpatton:
https://github.com/infor-design/enterprise-ng/blob/81fb2ba03161ad40cb46fbba258ce91381b21ca6/projects/ids-enterprise-ng/src/lib/button/soho-button.component.ts#L288-L298

Button also seems to be messaging the enterprise component appropriately.

pwpatton:
All the context menus seem to get destroyed when navigating away. 
is there a enterprise-ng sample that can be used to duplicate this?

prkarlsson:
We are on enterprise-ng version 4.7.x. If it is already fixed in later releases we'll just patch it temporarily in our local files until we upgrade to a version that includes the fix.

tmcconechy:
Ok, Closing unless we find additional information.

PR #200, opened 2018-09-28
Context menu breaks after toggling the element DOM

### LINKED ISSUE #84: Context menu breaks after toggling the element DOM
**Description**
When an element using soho-context-menu is toggled using *ngIf, the connected soho-popupmenu is destroyed and removed from DOM when the soho-context-menu is destroyed. This causes issues when several elements share the same soho-popupmenu, or if the toggled element is later shown.

**To Reproduce**
1. Edit \app\context-menu\context-menu.html sample to use *ngIf toggle
```
<div class="field">
    <label id="label1" name="label1" soho-label for="field1" >{{normalText}}</label>
    <input id="field1" name="field1" soho-context-menu menu="action-popupmenu" value="Right Click Here"
           *ngIf="!hideInput"
           (selected)="onSelected()"
           (beforeopen)="onBeforeOpen()"
           (close)="onClose()"
           (open)="onOpen()" />
  </div>
  <button soho-button="tertiary" (click)="hideInput = !hideInput">Toggle Input field</button>
```
2. Run the sample
3. Right-click the various elements that share the same soho-popupmenu element (works fine)
4. Remove input field by clicking the Toggle Input field button
5. Re-add input field by clicking the Toggle Input field button again
6. Right-click any of the elements that use the soho-popupmenu

**Expected behavior**
Popup-menu is shown on right-click

**Actual behavior**
Popup-menu is not shown. If right-clicking the toggled element, error is shown in console.

**Screenshots**
Attached a GIF showing the issue in sample

**Additional context**
I think sharing a menu between several different items is a common scenario, and this is no longer possible if any of the elements is removed from DOM (temporarily or not). In Homepages Menu widget, all the links / folders share the same context popup-menu, and when navigating between different links, context menu can no longer be used.

The issue was introduced when properly destroying the popup menu in context-menu-directive.ts.
https://github.com/infor-design/enterprise-ng/commit/9ce090817f961043cf8fb806d732ab7315b9d2b0#diff-864e73c329d001149d6cca6689f37fc4

(it used to set it to null only, but is now fully destroyed, unwrapped and removed from DOM with        this.contextMenu.destroy() )

**Possible solution**
Would it be possible to have an @Input to soho-context-menu, `unwrapOnDestroy` or similar, that defaults to false? If so, do not destroy and remove connected popupmenu from DOM.

toggle-menu

bthharper:

See #84 for a description of the issue.

Closes #84.

See related issue.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -20,12 +20,9 @@
         "request": "launch",
         "url": "http://localhost:9876/debug.html",
         "webRoot": "${workspaceFolder}",
-        "sourceMapPathOverrides": {
-          "webpack:/./*": "${webRoot}/*",
-          "webpack:/src/*": "${webRoot}/src/*",
-          "webpack:/*": "*",
-          "webpack:/./~/*": "${webRoot}/node_modules/*"
-        }
+        "sourceMaps": true,
+        "disableNetworkCache": true
+
       },
       {
         "name": "ng e2e",
--- a/angular.json
+++ b/angular.json
@@ -13,7 +13,7 @@
         "build": {
           "builder": "@angular-devkit/build-angular:browser",
           "options": {
-            "vendorSourceMap": false,
+            "vendorSourceMap": true,
             "sourceMap": true,
             "outputPath": "dist/ids-enterprise-ng-app",
             "index": "src/index.html",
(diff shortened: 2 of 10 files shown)


pwpatton:
other related issue might be #185

tmcconechy:
We have some failures here https://travis-ci.com/infor-design/enterprise-ng/builds/86300169#L711

bthharper:
Yes - same old random failure on datagrid.  Not sure what's up with that one - must be a timing issue I suspect.

Need to discuss this PR anyway - so will hold off fixing until we get a consensus on this fix.

PR #233, opened 2018-10-24
Missing options on menu button & ngZone

### LINKED ISSUE #226: `soho-menu-button` does not use ngZone
**Describe the bug**

All the components need updating to use ngZone to avoid redundant change detection being fired.

**To Reproduce**

Run the menu-button demo.
```ng s```

**Expected behavior**

Change Detection should not be fired multiple times when calling / handling any of the widget methods / triggers.

**Screenshots**

**Platform**

**Additional context**

### PR DESCRIPTION
**Pull request by bthharper**

Added several missing options from `soho-menu-button`, specifically `showArrow`.

Also upgraded the component to use ngZone.

Closes #226.

The sample code has been extended to include the `showArrow` option.

A new unit test has been added for the new options.

### DIFF
--- a/projects/ids-enterprise-ng/src/lib/menu-button/soho-menu-button.component.spec.ts
+++ b/projects/ids-enterprise-ng/src/lib/menu-button/soho-menu-button.component.spec.ts
@@ -212,7 +212,6 @@ describe('Soho Menu Button Unit Tests', () => {
   });
 });
 
-
 @Component({
   template: `<button soho-menu-button icon="user" menu="action-popupmenu"></button>
 <ul soho-popupmenu id="action-popupmenu">

### CODE REVIEW

File: projects/ids-enterprise-ng/src/lib/menu-button/soho-menu-button.component.ts [hunk 1]
```diff
@@ -49,62 +60,180 @@ export class SohoMenuButtonComponent implements AfterViewInit, OnDestroy {
   /** The icon to be used. */
   @Input() icon: string;
 
+  @Input() set autoFocus(value: boolean) {
+    this.options.autoFocus = value;
+    if (this.menuButton) {
+      this.menuButton.settings.autoFocus = value;
+      this.markForRefresh();
```
pwpatton: I don't believe you have to markForCheck for inputs. When an input is used this component should automatically be `markedForCheck`. The reason to call it manually is if you do something that needs to be reflected in the view but isn't going to automatically mark it.

pwpatton: Although I don't think this hurts anything.

Issue #239, opened 2018-10-25
Datagrid Examples are throwing errors or not loading.


tmcconechy:
**Describe the bug**

Noticed some of the datagrid examples are not loading or through errors.

**To Reproduce**
1. Click on each of the datagrid - examples
2. Some load the grid and some dont.

Problem one - Its not loading
http://localhost:4200/datagrid-groupedheader
- click into the form and the grid tries to load

Problem two - an error such as 
`Cannot read property 'selectedRows' of null`

**Additional context**
I think for one problem https://github.com/infor-design/enterprise-ng/blob/master/projects/ids-enterprise-ng/src/lib/datagrid/soho-datagrid.component.ts#L1383 should be `this.selectedRows` and also other places.

PR #241, opened 2018-10-25
fixes for datagrids not displaying after soho-datagrid ngZone refactor [v4.12.x]

pwpatton:
fix for #239

Open demo app and run each of the datagrid samples.

### DIFF
--- a/src/app/datagrid/datagrid-angular-editor.demo.ts
+++ b/src/app/datagrid/datagrid-angular-editor.demo.ts
@@ -1,17 +1,11 @@
 import {
   Component,
-  ContentChild,
   ViewChild,
-  AfterViewInit,
-  ComponentRef,
-  ElementRef,
-  Input,
   Inject,
-  OnDestroy
+  OnInit
 } from '@angular/core';
 
 import {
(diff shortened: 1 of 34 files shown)


src/app/datagrid/datagrid-angular-editor.demo.ts
```diff
@@ -304,16 +304,18 @@ export class DataGridAngularEditorDemoComponent implements AfterViewInit {
...
-      dataset: EDITORS_DATA,
-      selectable: 'single',
-      idProperty: 'productId',
-      stretchColumn: 'last',
-      editable: true,
-      isList: true,
-      filterable: true
-    };
+    // in ngAfterViewInit/Checked must set inputs from a timeout so that change detection is executed.
+    setTimeout(() => {
```
tmcconechy: So i guess this is a breaking change. Is there any way to not have to do this?

pwpatton: I don't think so, most folks are already doing this probably. It's surprising it worked at all w/o a timeout setting an input when in ngAfterviewInit and checked.

Issue #254, opened 2018-11-02
Update Hierarchy NG to match changes made to hierarchy enterprise. 


krishollenbeck:
**Describe the solution you'd like**
Hierarchy (enterprise) control now supports a submenu and disabled menu items in the actions menu. Need to add an example of that functionality to the NG version. Also need to update `SohoHierarchyAction` with new properties supported in the enterprise version.

**Additional context**
![hierarchy-example](https://user-images.githubusercontent.com/1056684/47925919-47228600-de8d-11e8-8f97-4b21ea41cb72.png)

https://github.com/infor-design/enterprise/issues/1044

tmcconechy:
This has PR - so should be fixed now 👍

Issue #214, opened 2018-10-12
Wrap renderLoop with Angular Component


krishollenbeck:
**Is your feature request related to a problem? Please describe.**
`renderLoop` appears to trigger a handful of the setTimeout violations we are seeing in chrome. Seems like this is a potential performance bottleneck.

**Describe the solution you'd like**
Currently `renderLoop` is initialized via Sohoxi.js. If we wrap this with an Angular wrapper (possibly a service), we could initialize it via  app.module. For example..

```
SohoRenderLoopService,
    {
      provide:    APP_INITIALIZER,
      useFactory: ... ,
      deps:       [ ]
    },
```

And then in the service have something like..

```
  this.ngZone.runOutsideAngular(() => {
      renderLoop()
  });
```

This way we can ensure `requestAnimationFrame` is running outside of Angular Zone.

tmcconechy:
Also should the loop only be running when a component needs it? For example if a toast is running

pwpatton:
Does this get started just due to sohoxi.js being loaded ? or is there function ep-ng calls where we could run outside angular?

tmcconechy:
Are you going to make this change @krishollenbeck ?

pwpatton:
One thing to not is that the renderLoop is called from an IIFE. So some sort or refactor will be required in soho/ep.
```
  // Setup a single instance of RenderLoop for export.
  var renderLoop = new RenderLoop();
  renderLoop.start();
```
which starts to `tick()` right away.

krishollenbeck:
@tmcconechy I am okay with working on this. @EdwardCoyle Is there a use case I can refer to test that it works correctly, or the way you intend it to work?

bthharper:
Is this complete now?

EdwardCoyle:
@bthharper yes, this is complete.  @krishollenbeck added the Angular part in #259, and infor-design/enterprise#1095 added the EP part.

PR #259, opened 2018-11-07
Fix for issues 254 & 214

krishollenbeck:
Fixes for the following tickets

**Hierarchy Issue:**
> Hierarchy (enterprise) control now supports a submenu and disabled menu items in the actions menu. Need to add an example of that functionality to the NG version. Also need to update SohoHierarchyAction with new properties supported in the enterprise version.

**RenderLoop Issue:**
> Currently renderLoop is initialized via Sohoxi.js. If we wrap this with an Angular wrapper (possibly a service), we could initialize it via app.module. For example..

1. https://github.com/infor-design/enterprise-ng/issues/254
2. https://github.com/infor-design/enterprise-ng/issues/214

**Hierarchy Issue:**
1. Apply changes from PR
2. Navigate to localhost:4203/hierarchy
3. Click the actions menu button on "Patricia Clark"
4. There should be a disabled "Change Pay Rate" and "Options" should have a sub menu

**RenderLoop Issue:**
1. Apply changes from PR
2. Navigate to the home page (or any page)
3. At the top of the page the current RenderLoop count should display on load.
4. Click stop to stop the loop & start to start the loop.
5. Comment out `this.renderLoop.start();` in **app.component.ts**. Notice the loop will not start now. This allows the Angular app to control when the loop should be initialized.


src/app/app.component.ts
```diff
@@ -1,10 +1,12 @@
 import {
+  AfterViewChecked,
   Component,
-  HostBinding,
+  HostBinding, OnInit,
```
pwpatton: put on own line

src/index.html
```diff
@@ -5,6 +5,11 @@
   <title>SohoXI - Angular Components</title>
   <base href="/">
 
+  <!-- prevent render loop from initializing on load -->
```
bthharper: Is this something we should all look at doing?

krishollenbeck: Yeah, before [Eds change](https://github.com/infor-design/enterprise/pull/1098) renderLoop initialized on load. He has now changed that to be optional via a config setting.

`if (this.settings.noAutoStart !== true) {
      this.start();
    }
`

We can set this in index.html before sohoxi.js loads . And then start it later manually in our ### app.component _ngOnInit()_ There is a requestAnimationFrame in the renderLoop that was causing changeDetection to fire. The renderLoop service allows us to call that outside of the Angular zone.

krishollenbeck: @bthharper I can add a note to the change log. Probably should have done that.

pwpatton: I think this can wait for ep-ng of 5.x.

bthharper: Okay - might be worth creating the new CHANGELOG section.

Issue #274, opened 2018-11-15
Add new features to Ng for v4.1.3


tmcconechy:
**Is your feature request related to a problem? Please describe.**

Go through change log for 4.124 in ids enterprise https://github.com/infor-design/enterprise/blob/4.14.x/docs/CHANGELOG.md

Add any new settings or make issues for the more complicated things that cannot be hit on first pass (which will be listed here)

- [x] #1088 - Added a new config option to menu button `hideMenuArrow`
- [x] #269 - Flex toolbar wrapper
- [x] #862 - Removed customVaidation setting.
- [x] #214 - Stopping renderloop
- [x] #1094 - Bar chart short, medium label support for adapting the chart to responsive views.
- [x] #1046 - Added a maxLength option to texarea
- [x] #1147 - Added a maxGrow option to texarea
- [x] #1170 - Added customColors option to the color picker
- [x] #1003 - Tree got aded sortstop and sortstart events 
- [x] #398 - Added beforeOpen to autocomplete.
- [x] #1172 - Add/confirm Promises polyfill

tjamesallen15:
Question on 1094. The settings are add inside of the "ticks" which is already added in the angular. I don't see the need of change. Open for suggestions here. Thank you

tmcconechy:
You can test this example out for 1094 in NG. https://github.com/infor-design/enterprise/blob/master/app/views/components/bar/example-long-text.html#L23-L25

I guess you might not need to do anything actually as the data in in the model code https://github.com/infor-design/enterprise-ng/blob/master/src/app/bar/bar.demo.ts#L20

Maybe just update that example to show the different label sizes (then resize the page like in EP to see them in action).

Thanks!

tjamesallen15:
Thanks for the reply about 1094 @tmcconechy.

Another question is how should I apply 1172 in NG? I might be confused on what should be applied that's why I'm asking for clarification.

Thanks !

tjamesallen15:
I have an issue also in my workspace. It doesn't contain any errors when building "ng s".

However the view cannot be displayed on the browser.

<img width="1010" alt="screen shot 2019-01-08 at 11 15 37 am" src="https://user-images.githubusercontent.com/34049950/50807853-e6a14080-1336-11e9-84e1-fb3dc16e8e8b.png">

Thanks

tmcconechy:
Looks like your missing the ids-enterprise code or it could be the `--serve-path` change recently.
Try:

```
npm install
npm run build
[...]

PR #280, opened 2018-11-27
Added a new config option to menu button `hideMenuArrow`

bthharper:

Add the new configuration option to allow a menu button's icon to be hidden.

Also includes a small version bump for angular.

#274 

Build the project, and then viewing the menu-button demo.  There will be a user icon without an arrow.


projects/ids-enterprise-ng/src/lib/button/soho-button.component.ts
```diff
@@ -96,10 +96,22 @@ export class SohoButtonComponent implements AfterViewInit, OnDestroy, OnInit {
     this._buttonOptions.replaceText = replaceText;
     if (this.jQueryElement) {
       this.button.settings.replaceText = replaceText;
-      // todo: how to update the button replaceText changes?
+      // todo: how to update the button when replaceText changes?
```
tmcconechy: We probably should make this into a function https://github.com/infor-design/enterprise/blob/master/src/components/button/button.js#L67-L71 that can be called in the NG component

projects/ids-enterprise-ng/src/lib/button/soho-button.component.ts
```diff
@@ -96,10 +96,22 @@ export class SohoButtonComponent implements AfterViewInit, OnDestroy, OnInit {
...
     }
   }
 
+  @Input() set hideMenuArrow(value: boolean) {
+    this._buttonOptions.hideMenuArrow = value;
+    if (this.button) {
+      this.button.settings.hideMenuArrow = value;
+      // todo: how to update the button when hideMenuArrow changes?
```
tmcconechy: Same here but there is only a remove no way to re-add so need a function

projects/ids-enterprise-ng/src/lib/button/soho-button.component.ts
bthharper: Will raise another ticket for this ... on enterprise.

PR #289, opened 2018-12-11
Use npx in package scripts

nickwynja:
Use local managed version of ng cli instead of global.

1. Pull branch
2. `npm run build` works without errors


.travis.yml
```diff
@@ -7,7 +7,7 @@ cache:
 install:
   - npm install
 before_script:
-  - npm install -g @angular/cli codecov
+  - npm install -g codecov
```
bthharper: Installing the global cli give you the commands on the path, do we want to remove that?

docs/DEVELOPER.md
```diff
@@ -10,19 +10,17 @@
 
 - Clone The main repo fx `git clone https://github.com/infor-design/enterprise-ng.git`
 - Open a command prompt to enterprise-ng
-- Type `npm install -g @angular/cli@latest`
```
bthharper: I would leave this, but use 7.1.x.

PR #306, opened 2018-12-19
version change to projects/package.json so that the scripts will bump…

pwpatton:
publish scripts need `<version>-dev` in package.json version property to publish a dev build


pwpatton:
Ok, I've included the main package.json and the svg.html files that were updated due to running `npm run update-enterprise`

PR #313, opened 2019-01-03
274 - Added beforeOpen config in AutoComplete

tjamesallen15:

Added beforeComplete in AutoComplete 

Partial Work for #274

### DIFF
--- a/projects/ids-enterprise-ng/src/lib/autocomplete/soho-autocomplete.component.ts
+++ b/projects/ids-enterprise-ng/src/lib/autocomplete/soho-autocomplete.component.ts
@@ -77,6 +77,8 @@ export class SohoAutoCompleteComponent extends BaseControlValueAccessor<string> 
   @Output() change: EventEmitter<SohoInputEvent[]> = new EventEmitter<SohoInputEvent[]>();
   @Output() selected: EventEmitter<Object[]> = new EventEmitter<Object[]>();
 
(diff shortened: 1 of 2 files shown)


tjamesallen15:
Same with this one. I'm not so sure with the changes. Open for suggestions and ideas. Thank you!

projects/ids-enterprise-ng/src/lib/autocomplete/soho-autocomplete.component.ts
```diff
@@ -147,7 +149,8 @@ export class SohoAutoCompleteComponent extends BaseControlValueAccessor<string>
     // Bind to jQueryElement's events
     this.jQueryElement
       .on('selected', (...args) => this.selected.emit(args))
-      .on('change', (e: any, args: any[]) => this.onChange(args));
+      .on('change', (e: any, args: any[]) => this.onChange(args))
+      .on('beforeopen', (e: any, args: SohoAutoCompleteEvent) => this.beforeopen.emit(args));
```
pwpatton: @bthharper  do we want to have folks convert to running ep code outside ngZone?

bthharper: @pwpatton  - I assume none of autocomplete has been updated to correctly switch zones yet, but it might be a good idea to do that here.  I would have expected the event handler to swap back into the Angular zone.

bthharper: Maybe also add some types to the change event.

Issue #318, opened 2019-01-10
Datagrid Tree, Have to hit enter twice to filter using server side filtering


krishollenbeck:
**Describe the bug**
Have to hit enter twice to filter using server side filtering. This is caused by setting activePage to -1 in the angular component wrapper. _soho-datagrid.component.ts_ I believe this was a workaround for a past issue. It doesn't appear to be needed anymore.

```
if (this.treeGrid) {
    pagerInfo.activePage = -1;
}
```

Removing this fixes the issue.

**Additional Info**
The reason this happens is because _sohoxi.js_ calls `renderHeaders()` from within `loadData: function`. 

```
 if (pagerInfo.type === 'initial') {
        this.clearHeaderCache();
        this.restoreUserSettings();
        this.renderRows();
        this.renderHeader();
  }
```

This only happens if _pagerInfo.type_ is initial. Treegrid never gets set to initial because _activePage_ is -1. `renderHeader()` then calls `applyFilter()` which preforms and initial filter causing a flag called `restoreFilterClientSide` to be set to false. This never happens for treegrid. Ultimately this causes treegrod to exit on the first "enter" or the initial filter when it hits this code within `applyFilter().`

```
if (this.restoreFilterClientSide) {
        this.restoreFilterClientSide = false;
        return;
}
```

This causes a return to occur before:

 `this.element.trigger('filtered', { op: 'apply', conditions: conditions, trigger: trigger });`

**To Reproduce**
1. Get code from PR https://github.com/infor-design/enterprise-ng/pull/319
2. Navigate to http://localhost:4200/ids-enterprise-ng-demo/datagrid-treegrid-dynamicfilter
3. Toggle filter row
4. Apply this code back to soho-datagrid.component.ts
```
if (this.treeGrid) {
    pagerInfo.activePage = -1;
}
```
5. Filter the list. Notice enter needs to be hit twice. Remove above code and it should filter as expected.

**Expected behavior**
Should only have to hit enter 1 time to filter.

**Version**
- ids-enterprise-ng: 4.12.1 through 5.1.0-dev.20190109

PR #319, opened 2019-01-10
318 - Fix having to hit enter twice to filter in datagrid

krishollenbeck:
- Updated example to illustrate the issue. Used HTTP service to illustrate how datagrid tree is used in a real world scenario.

Have to hit enter twice to filter using server side filtering. This is caused by setting activePage to -1 in the angular component wrapper. _soho-datagrid.component.ts_ I believe this was a workaround for a past issue. It doesn't appear to be needed anymore.

https://github.com/infor-design/enterprise-ng/issues/318

**To Reproduce**
1. Get code from PR
2. Navigate to http://localhost:4200/ids-enterprise-ng-demo/datagrid-treegrid-dynamicfilter
3. Toggle filter row
4. Apply this code back to soho-datagrid.component.ts
```
if (this.treeGrid) {
    pagerInfo.activePage = -1;
}
```
5. Filter the list. Notice enter needs to be hit twice. Remove above code and it should filter as expected.

### DIFF
--- a/src/app/datagrid/datagrid-treegrid-dynamicfiltering.demo.html
+++ b/src/app/datagrid/datagrid-treegrid-dynamicfiltering.demo.html
@@ -12,11 +12,13 @@
   <div class="row top-padding">
     <div class="twelve columns demo" role="main">
       <section class="scrollable contained-scrolling-right-bottom" style="height: 100%">
-        <div soho-busyindicator soho-datagrid
-             selectable="single"
+        <div soho-busyindicator soho-datagrid selectable="single"
+
+             *ngIf="gridOptions"
              [treeGrid]="true"
-             [columns]="columns"
+             [gridOptions]="gridOptions"
              [dataset]="data"
+
              (filtered)="onFiltered($event)">
         </div>
       </section>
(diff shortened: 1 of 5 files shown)


src/app/datagrid/datagrid-treegrid-dynamicfiltering.demo.html
```diff
@@ -12,11 +12,13 @@ <h2 class="fieldset-title">Soho Data Grid - TreeGrid Dynamic Filter</h2>
...
     <div class="twelve columns demo" role="main">
       <section class="scrollable contained-scrolling-right-bottom" style="height: 100%">
-        <div soho-busyindicator soho-datagrid
-             selectable="single"
+        <div soho-busyindicator soho-datagrid selectable="single"
+
+             *ngIf="gridOptions"
              [treeGrid]="true"
-             [columns]="columns"
+             [gridOptions]="gridOptions"
```
tmcconechy: Whats the explanation about this change? I'm wondering why you dont need columns. But have a new gridOptions rather than just using the settings object that matches EP?

Probably a reason im sure but just wondering as it seems inconsistent with the EP API

tmcconechy: Ok, I see its just part of the sample, i thought it was API. So i think this matters less.

PR #371, opened 2019-02-22
336 - Added a couple toolbar missing settings

### LINKED ISSUE #336: Introduce ResetLayout to Toolbar for Datagrid
We need to have a way to reset to default the column order on the DataGrid

Same functionality as it is today in the resetLayout in the non-angular solution. When the user moves the columns and clicks reset to default in the toolbar, then the datagrid resets the columns to the initial order.

**Additional context**

tmcconechy:

Checked datagrid toolbar options and added a few missing ones.

Closes #336


pwpatton:
Hi Tim, we could collaborate on new settings you add in the d.ts and I could make the @INputs and add to the spec.tests for each.

tmcconechy:
We didnt ever have inputs for these - no.

PR #375, opened 2019-02-25
374 - Added safety check

### LINKED ISSUE #374: Datagrid: Console error when running clearFilter
**Describe the bug**
One of our testers often get an issue with a console log _"[ERROR] Exception in core.raise TypeError: Unable to get property 'clearFilter' of undefined or null reference" or "ERROR TypeError: Cannot read property 'clearFilter' of null"._ that shows up in Homepages Administration. I cannot reproduce it, but I have investigated it and found that it comes from the clearFilter method in soho-datagrid.component.ts where this.datagrid is undefined: 
```
  /**
   * Clears any filter defined, for this datagrid.
   */
  clearFilter(): void {
    this.ngZone.runOutsideAngular(() => {
      this.datagrid.clearFilter();
    });
  }
```

So my question is if you can please add a check to see if this.datagrid is null before calling clearFilter? We already have this test in Homepages before calling this method. Perhaps it's a timing issue so that it's not null when we call the ClearFilter method, but sometimes its null when runOutsideAngular is executed.

**To Reproduce**
Not easily reproducible :(

**Expected behavior**
It would be good if the method had a null check so it doesn't throw an error if this.datagrid is null/undefined when the clearfilter method should be executed
```
	clearFilter(): void {
		this.ngZone.runOutsideAngular(() => {
			if (this.datagrid) {
				this.datagrid.clearFilter();
			}
		});
	}
```

**Version**
- ids-enterprise-ng: 5.0

**Screenshots**
image

**Platform**
 - OS Version: Windows 10
 - Browser Name: chrome / edge
 - Browser Version: Version 72.0.3626.109 (Official Build) (64-bit)

**Additional context**

Thanks!

tmcconechy:
Added a check as mentioned on #374 to prevent a random error.

Closes #374 

- NA - Not easily reproducible

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,7 @@
 - `[DataGrid]` 4.0 Datagrid expose error row clear functions `PWP` ([#314](https://github.com/infor-design/enterprise-ng/issues/314))
 - `[DataGrid]` Added a few missing toolbar options to the types. `TJM` ([#336](https://github.com/infor-design/enterprise-ng/issues/336))
 - `[DataGrid]` Added headerAlign option to datagrid column. `PWP` ([#304](https://github.com/infor-design/enterprise-ng/issues/304))
+- `[DataGrid]` Added safety check on clearFilter for random errors. `TJM` ([#374](https://github.com/infor-design/enterprise-ng/issues/374))
 - `[DatePicker]` Added firstDayOfWeek option to date picker component `PWP` ([362](https://github.com/infor-design/enterprise-ng/issues/362)'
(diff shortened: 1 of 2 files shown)


pwpatton:
Due to ngZone, order of operation, and implementing usage these things can be very hard to find and are usually implementation specific. A safety check is the standard way of making sure it at least doesn't crash in the ng wrapper component. If the behavior is still not working for the implementer it's usually an issue with their implementing code expecting the component wrapper to be all set up before it actually is.

SofiK:
> Due to ngZone, order of operation, and implementing usage these things can be very hard to find and are usually implementation specific. A safety check is the standard way of making sure it at least doesn't crash in the ng wrapper component. If the behavior is still not working for the implementer it's usually an issue with their implementing code expecting the component wrapper to be all set up before it actually is.

Yes, for us it's the other way around. We use it when it's about to be removed, but it doesn't know that it's about to be removed and is then removed and crashes :)

Thanks for the fix!

PR #438, opened 2019-04-17
Added version number.

### LINKED ISSUE #429: Add `ids-enterprise-ng` version to page body

To help determine the **ids-enterprise-ng** version an application is compiled with, the current version should be added to the page body, like the **ids-enterprise** version currently is.

The result should be as follows:

```html
<html data-ids-enterprise-ng-version="5.2.1" data-sohoxi-version="4.17.1" ...>
...
</html>
```

The implementation for this could be done as an **APP_INITIALIZER** that modifies the body of the application on start.

**Additional context**

### LINKED ISSUE #432: Accessibility - Modals Have the Wrong Role
**Describe the bug**
The modal in NG has the wrong role.

**To Reproduce**
1. Go to http://localhost:4200/modal 
2. Click on modal to open
3. Inspect the DOM
4. You will see `role="alertdialog"` this is only valid for messages not on modals. 
5. http://latest-enterprise.demo.design.infor.com/components/modal/example-index.html

**Expected behavior**
Should have the correct role in this case. Something is wrong in the wrapper.

- [x] Add an Axe check for correct accessibility roles - this isnt possible.  - The role is valid but the context is wrong.

**See also**
https://jira.lawson.com/browse/LMCLIENT-24206

bthharper:

To help determine the **ids-enterprise-ng** version an application is compiled with, the current version is now added to the html element (if present).

Closes #429.
Closes #432 

clone the enterprise-ng library

```sh
npm i
npm run build:lib
npm run start
```

Verify that the html element has the correct version number.


projects/ids-enterprise-ng/src/lib/version/version-initializer.module.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { SohoVersionInitializerModule } from './version-initializer.module';
+
+describe('SunSystemsAuthModule', () => {
```
tmcconechy:
```suggestion
describe('Soho Authorization Module', () => {
```

projects/ids-enterprise-ng/src/lib/version/version-initializer.service.spec.ts
```diff
@@ -0,0 +1,17 @@
+import { TestBed } from '@angular/core/testing';
+
+import { SohoVersionInitializerService } from './version-initializer.service';
+
+describe('SunSystemsI18nInitialiserService', () => {
```
tmcconechy:
```suggestion
describe('Soho I18n Initialiser Service', () => {
```

src/app/modal-dialog/modal-dialog.demo.ts
```diff
@@ -30,7 +30,7 @@ export class ModalDialogDemoComponent {
   public closeResult: string;
 
   public title = 'Example Modal Dialog';
-  public isAlert = true;
+  public isAlert = false;
```
tmcconechy: https://github.com/infor-design/enterprise-ng/issues/432#issuecomment-483993245 - Does tihs solve this issue or do we need to update the demo as well. If so maybe just update the demo as well.

projects/ids-enterprise-ng/src/lib/version/version-initializer.service.ts
```diff
@@ -0,0 +1,29 @@
+import { Injectable } from '@angular/core';
+
+// @todo unable to load the version from a file not included as a module in the package.
+// const version = require('../../../../package.json').version;
+// const version = process.env.npm_package_version;
+const version = '5.2.1';
```
tmcconechy: Seems to be a few suggestions on https://github.com/angular/angular-cli/issues/11931 you could try

package.json
```diff
@@ -28,19 +28,19 @@
...
-    "@angular/router": "~7.2.9",
+    "@angular/animations": "~7.2.12",
+    "@angular/common": "~7.2.12",
+    "@angular/compiler": "~7.2.12",
+    "@angular/core": "~7.2.12",
+    "@angular/forms": "~7.2.12",
+    "@angular/http": "~7.2.12",
+    "@angular/platform-browser": "~7.2.12",
+    "@angular/platform-browser-dynamic": "~7.2.12",
+    "@angular/router": "~7.2.12",
```
tmcconechy: Seems like latest is 
```
usmvtmcconechy:enterprise-ng tmcconechy$ npx ng update
    We analyzed your package.json, there are some packages to update:

      Name                               Version                  Command to update
     --------------------------------------------------------------------------------
      @angular/cli                       7.3.6 -> 7.3.8           ng update @angular/cli
@angular/core                      7.2.10 -> 7.2.13         ng update @angular/core
```

src/app/modal-dialog/modal-dialog.demo.ts
bthharper: Yes - this solves the issue in the demo.

tmcconechy:
@btharrper I’m not sure if this should just go as is or if you want me to request changes. I could let it go and add the updates later on those two packages. And if you think the hardcoding of the version has no solution?

bthharper:
@tmcconechy  - my plan is to add a `version.json` file which can be updated by the release scripts, otherwise things may get out of sync.

clepore:
Also @bthharper you should be able to checkout a branch right from the repo, you shouldn't need to use a fork if you don't want to 👍 (up to you).

PR #450, opened 2019-04-25
updating ep version to 4.19.0-dev.20190424

pwpatton:
To get the latest changes from EP that includes some new theming abilities.


projects/ids-enterprise-ng/package.json
```diff
@@ -1,17 +1,17 @@
...
     "@types/d3": "4.13.0",
-    "ids-enterprise": "4.18.0-dev.20190409",
+    "ids-enterprise": "4.19.0-dev.20190424",
     "jquery": "3.3.1",
     "d3": "4.13.0"
   },
   "peerDependencies": {
-    "@angular/common": "^7.2.0",
-    "@angular/core": "^7.2.0"
+    "@angular/common": "^7.1.0",
```
tmcconechy: This seems strange to me as you moved it back? But actually its 7.2.9 now in the main package.json https://github.com/infor-design/enterprise-ng/blob/master/package.json#L31

pwpatton: I've moved this back to 7.2 do we want to move it to 7.2.9?

PR #452, opened 2019-04-25
Updating some travis-ci settings

clepore:
I updated that travis-ci settings so that
- "Function" tests and "Lint" tests will now run in parallel (saves a bit of time)
- Added an `.nvmrc` to specify the support NodeJS version. This is also helpful as a dev so you can easily switch to the supported version in each repo by doing `nvm use`.

n/a

1. Check out the builds and see that they are parallel stages:
    Screen Shot 2019-04-25 at 5 19 00 PM
1. Test `nvm use`

---
🚫 Update the changelog (if needed)
🚫 Check back to make sure tests pass on Travis


tmcconechy:
We still going to run both branch and PR here on this one or change it here too?

PR #456, opened 2019-04-29
2044 - Reverts previous fix and fixes a dependency

tmcconechy:

Reverted the last fix and did it in the EP side. https://github.com/infor-design/enterprise/pull/2046
Added some comments for a situation that may occur using inputs.

Once 4.19 is merged we should put that version on 

#2044 

- Build this branch https://github.com/infor-design/enterprise/pull/2046
- copy over to the NG project or use npm link
- open http://localhost:4200
- make sure the menu can be toggled


src/app/app.component.ts
```diff
@@ -49,10 +49,17 @@ export class AppComponent implements AfterViewInit {
   }
 
   ngAfterViewInit(): void {
+
+    /**
+    * Note: If using an input like [triggers]="[ '.application-menu-trigger' ]"
+    * hookup the app menu trigger once the afterViewInit is called. This will
+    * ensure that the toolbar has had a chance to create the application-menu-trugger
+    * button.
+    * this.applicationMenu.triggers = [ '.application-menu-trigger' ];
```
pwpatton: Probably want this in a time out so it can complete a tick.

And the next set of code in another timeout so it can rely on the button being hooked up

tmcconechy: I tested it and it works fine, but its just a comment now, as its not the default way to do it.

tmcconechy: Thought maybe was a good place to put it  but maybe we dont  need?

pwpatton: Ok, what is the default way to do it?

tmcconechy: Without adding the triggers input inline on the component.

tmcconechy:
@pwpatton and @fitzorama please approve again? I bumped the versions so it definitely will show as working.

PR #458, opened 2019-05-01
Added `soho-icons-ext` back.

### LINKED ISSUE #436: Breaking change with soho-icons-ext in 5.2.1 comparing to 5.1.1 shouldn't be in minor release
**Describe the bug**
Upgrading to 5.2.1 from 5.1.1 resolves in failed build due to `soho-icons-ext` no longer exists. This is a breaking change and it shouldn't be introduced in minor release according to sem version convention.

**To Reproduce**
1. Use an app with 5.1.1 `ids-enterprise-ng`
2. Declare `<soho-icons-ext></soho-icons-ext>` in `app.component.html`
3. Upgrade `ids-enterprise-ng` to 5.2.1
4. Start dev server (or run prod build) and see errors (IDE will notice as well).

**Expected behavior**
`soho-icons-ext` shouldn't be removed in a minor release version. Removing this component is a breaking change and it should be in major release version.

**Version**
- ids-enterprise-ng: 5.2.1

**Screenshots**
**N.A.**

**Platform**
 - Device **N.A.**
 - OS Version: Windows 10
 - Browser Name **[e.g. chrome, safari, stock browser]**
 - Browser Version **[e.g. 22, 66.0.3359.181 (Official Build) (64-bit)]**

**Additional context**
Does `ids-enterprise-ng` follow sem version convention ?

bthharper:

Upgrading to 5.2.1 from 5.1.1 resolves in failed build due to `soho-icons-ext ` no longer exists. This is a breaking change and it shouldn't be introduced in minor release according to sem version convention.

Closes #436.

With an app referencing 5.1.1 of `ids-enterprise-ng`
Declare <soho-icons-ext></soho-icons-ext> in `app.component.html`
Upgrade `ids-enterprise-ng` to 5.2.1
Run the server `npm run start` and see errors in the console.

<!--


pwpatton:
Agreed that I should not have removed this. But is soho-icons-ext essentially a NOOP? I.e. there is no actual svg definitions as a result of using this element?

clepore:
We left the file in there so that it would not break builds or references, but yes it does only contain a comment.

tmcconechy:
We probably CAN remove it but we would want to make a 6.0.0 maybe we can remove it for good when we do the bigger NG version updates.

bthharper:
Do it as part of the angular 8 upgrade.

PR #500, opened 2019-05-24
Set default locale to en-US

bthharper:

Current locale is en-GB (checked in by accident), which and this causes the themes to not be localised.  

N/A

Clone, compile and run.  Check the theme menu is displayed without [...]


tmcconechy:
Also related is PR https://github.com/infor-design/enterprise/pull/2238 i fixed locale so en-GB will show the translations for english correctly

PR #528, opened 2019-06-14
Events not fire without ngmodel

bthharper:

The change event is not fired when:
* a `soho-dropdown` not associated with an ng model is changed.
* a `soho-editor` is changed.

Also fixed the Soho


bthharper:
@tmcconechy - yes my changes are in there and the branch is from master.  However, there are also some additional changes caused by reformatting, removal of additional spaces, lining up code etc.  I might have activated format on save by accident, so wasn't intending these.

PR #530, opened 2019-06-20
526 - Fix datagrid code-block editor focus

### LINKED ISSUE #526: Datagrid:  Tabbing into the next visible column doesn't work from a code-block cell
Describe the bug
When tabbing out of a cell with a code-block, focus leaves the datagrid. It should put focus in the next visible cell. 

To Reproduce

1. Using the attached datagrid-code-block-editor.demo.ts
2. Put focus in the Company cell of any row
3. Hit the tab key to into the Code Block cell
4. Continue hitting the tab key to tab across the three input fields that make up the Code Block
5. Attempt to tab into the Name cell

Expected behavior
When tabbing out of a cell containing a code block, focus should be on the next visible cell.

Additional context
Changes to datagrid-code-block-editor.demo.ts
- Render only 3 fields and removed the more button for each field
- Grid options have changed to reflect what Landmark would use
- Changed the order of the columns to have a column after the code-block column

Changes to code-block.component.ts
- Removed the hidden icon-launch button

Note:
When the last field in the code-block has a more button AND the code-block-buttons div contains a button (see code-block.component.ts) then tabbing across the row works.

First cut of code-block implementation will not go out with a more button for each code block field.

deep7102:
Fixed code-block editor focus was not working with soho datagrid.

Closes #526

- After ([PR 2412](https://github.com/infor-design/enterprise/pull/2412)) merged
- Pull and build this branch
- Go to http://localhost:4200/ids-enterprise-ng-demo/datagrid-code-block-editor
- Put focus in the Company cell of any row
- Press tab key to into Code Block cell
- Continue pressing tab key to tab across the input fields that make up the Code Block
- Attempt to tab into the Name cell
- Focus should be working fine


src/app/code-block/code-block.component.ts
```diff
@@ -66,9 +68,11 @@ export class CodeBlockComponent implements OnDestroy {
   ngOnDestroy() {
   }
 
-  focus() {
+  focus($event?: any) {
```
fitzorama: I believe we have been trying to use JQuery.TriggeredEvent for the event type

src/app/datagrid/datagrid-code-block-editor.demo.ts
```diff
@@ -188,7 +198,7 @@ export class CodeBlockEditorComponent  implements SohoDataGridCellEditor {
   }
 
   focus() {
-    this.codeblock.focus();
+    this.codeblock.focus(event);
```
fitzorama: Not seeing the incoming event parameter on the method

deep7102:
Hi @fitzorama not familiar how to use event args here as you suggested `JQuery.TriggeredEvent`, saw that `event` var is looks coming with focus globally as keyboard event so I use type as `any`, but now its updated to `MouseEventInit`

src/app/code-block/code-block.component.ts
fitzorama: @deep7102 You will want something like this >
focus($event?: JQuery.TriggeredEvent) {

src/app/datagrid/datagrid-code-block-editor.demo.ts
fitzorama: @deep7102 Something like this >
focus(event: JQuery.TriggeredEvent) {

deep7102: Hi @fitzorama I did tried that did not work
- Add in to interface here `/projects/ids-enterprise-ng/src/lib/datagrid/soho-datagrid.d.ts` as
`focus(event: JQuery.TriggeredEvent) {`
- Then use here `/src/app/code-block/code-block.component.ts` as 
`focus(event: JQuery.TriggeredEvent) {`
- See it always undefined

Issue #553, opened 2019-07-25
Message.close() does not close the dialog


boehmd1:
**Describe the bug**
Message.close() does not close the dialog

**To Reproduce**
1. Goto --> http://localhost:4200/ids-enterprise-ng-demo/message
2. Click the button 'OpenAndCloseProgramtically' Example button

**Expected behavior**
Dialog is closed

PR #554, opened 2019-07-25
The close modal api wasn't working

boehmd1:
Calling the close method does not close the message dialog.

https://github.com/infor-design/enterprise-ng/issues/553

1. Goto --> http://localhost:4200/ids-enterprise-ng-demo/message
2. Click the button 'OpenAndCloseProgramtically' Example button

expect that the dialog closes in 3 seconds.


tmcconechy:
FWIW If you go to http://master-enterprise.demo.design.infor.com/components/message/example-index.html you can open a message and call `$('.modal').data('modal').close()`  in the console. This means somehow we have to call close on the underlying modal plugin as message uses this. (so you should be able to do it by some jQuery call on the component)

PR #571, opened 2019-08-08
Created a directive to allow homepage element to set it's height to the computed containerHeight

pwpatton:
W/o a height on homepage we weren't able to use position sticky on any element before the home page. With it's element height computed correctly the container now autoheights to the correct size and the sticky tabs we have stick at their correct position.

- open ng app
- H -> HomePage Scenario A
- verify in debugger that div.homepage is getting it's height reset on each resize of the homepage container.


projects/ids-enterprise-ng/src/lib/homepage/soho-homepage-sizer.directive.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import {
+  AfterViewInit,
+  Directive,
+  ElementRef, HostBinding,
+  NgZone, OnDestroy
+} from '@angular/core';
+
+/**
+ * Angular Wrapper for the SohoAlert Directive.
```
fitzorama: SohoSizer

tmcconechy:
Are we waiting on things here?

pwpatton:
yea, waiting for an EP build with the dependent homepage.js code.

tmcconechy:
OK i made a `4.21.0-dev.20190809` and pushed it on this branch. Give it a test...

PR #573, opened 2019-08-12
updating to latest ep dev version

pwpatton:
Created new dev package, updating lib package.json to new ep-ng version.


pwpatton:
@tmcconechy yea, I did that but only got the package.json fiel change and the build generated version.txt file change. So I only checked in the lib package.json since the other already had the correct ids-enterprise verison.

pwpatton:
@tmcconechy do we promote in the version.txt file? Or is that a build generated artifact solely.

PR #577, opened 2019-08-13
ng-zone'ing modal dialog

pwpatton:


pwpatton:
@bthharper what Error?

pwpatton:
ok, got that Error moved outside the runOutsideAngular block
Also fixed a memory l leak where modal event listeners are not removed.

PR #580, opened 2019-08-14
update ep version

pwpatton:
update ep version


tmcconechy:
Just FYI the change you wanted for Ng8 PR is not in until tonights build

pwpatton:
@tmcconechy I think that is fine. If we get an ep build tonight then we can make an ng build tomorrow

PR #624, opened 2019-09-30
Forcing a datagrid (in a tab) to use content scrollbars

bthharper:

 Example of a `soho-datagrid` hosted inside a `soho-tab` component in a header / detail pattern.

No specific issue, however this example may assist other developers in how to make this layout work with suitable scrollbars on the datagrid.

Checkout repository
```sh
npm i
npm run build:lib
npm run start
```

Navigate to http://localhost:4200/ids-enterprise-ng-demo/datagrid-tab.


bthharper:
> Were the modal dialog changes and the chang eto add a class to the html intentional? Of maybe got in from looking at other issues?

@tmcconechy  - I've removed the html change, was looking at the theme issue we discussed the other day and forgot to remove that.  I would like to keep the dialog change as that show a datagrid in a dialog using a scrollbar on the content.

Issue #636, opened 2019-10-07
Soho Modal Dialog Memory Leak


nbcp:
**Describe the bug**

**To Reproduce**
1. Go to http://localhost:4200/ids-enterprise-ng-demo/modal-dialog
2. Click Example button (Opens a modal)
3. Click Cancel / Submit (Closes the modal)
4. Generate a Heap Snapshot using Chrome DevTools' Memory tab
5. Filter objects using "modal" keyword

**Expected behavior**
It should not show references to these objects:
- Modal
- SohoModalDialogRef
- ExampleModalDialogComponent
- SohoModalDialogInjector

**Version**
- ids-enterprise-ng: latest

**Screenshots**
![image](https://user-images.githubusercontent.com/11013464/66302333-5591e100-e92b-11e9-818f-3da2e6981b13.png)

nbcp:
So far I have came up with a partial solution which I have pushed in **soho-modal-memleak** branch. It was only able to solve 3/4 of the leaking objects (see screenshot below). Part of the changes was to remove the injector being passed into createComponent because it creates a circular dependency (injector > dialogRef > component > injector). 

What remains is to figure out how to free up **ExampleModalDialogComponent** from memory. I'm stuck in here and already ran of ideas. Need everyone's help to be able to move forward and resolve this issue.

![image](https://user-images.githubusercontent.com/11013464/66304410-ab688800-e92f-11e9-8e32-f2ca1b81540e.png)

tmcconechy:
Yeah @nbcp i'm not 100% sure about this but i tried your branch and steps and still get all the same elements in the heap output.

I would think you should not have to do this for all variables in the object..

![Screen Shot 2019-10-07 at 11 11 01 AM](https://user-images.githubusercontent.com/814283/66324405-5870df80-e8f3-11e9-8992-1ebbc782a8ad.png)


```
import { Component, OnDestroy } from '@angular/core';

/**
 * This is an example of a simple dialog component, that can be instantiated
 * numerous times using the SohoModalDialogService.
 */
@Component({
  templateUrl: './example-modal-dialog.component.html'
})
export class ExampleModalDialogComponent implements OnDestroy {
   public model = {
    header: 'Default Header Text',
    comment: 'This task needs to be escalated to maximum priority and delivered by the end of this week.',
   };

  ngOnDestroy() {
    this.model = null;
  }
}
```

Maybe this garbage collects later or the problem is really because its not fixed in modal?

[...]

Issue #135, opened 2018-07-13
Contextual Action Panel is leaking divs when closed.


bthharper:
**Describe the bug**

When running the demo app for the Contextual Action Panel, closing the modal dialog does not remove all the added markup.  Adding this multiple times causes multiple `div` elements to be added to the page.

**To Reproduce**

Run the angular demo app and navigate to the "Contextual Action Panel" demo, and open and close the panel several times.

The standard modal dialog works correctly, cleaning up all dynamic markup.

**Expected behavior**

Any markup added as part of the CAP should be removed when closed.

**Screenshots**

![image](https://user-images.githubusercontent.com/8340092/42681349-543b2f1a-867f-11e8-969a-877672d1f5a7.png)

**Platform**

N/A

**Additional context**

tmcconechy:
Noting that this seems to be angular specific as it is not occurring here http://master-enterprise.demo.design.infor.com/components/contextualactionpanel/example-jquery.html or here http://master-enterprise.demo.design.infor.com/components/contextualactionpanel/example-index.html

tmcconechy:
I think this issue was fixed.

PR #639, opened 2019-10-10
Soho modal memleak

nbcp:
This has not yet fixed the remaining leaking angular component. 
From what I have observed so far, when the angular component is moved around the DOM (via jQuery methods or some native DOM manipulation methods) it breaks some references used in the angular component life cycle. That's why it was not unable to destroy itself properly using `ComponentRef.destroy()`.

Maybe the only thing that could fix this is to reimplement the whole SohoModalDialogModule and create all the modal's parts using angular components and just apply the css classes from enterprise.

#636 

1. Go to http://localhost:4200/ids-enterprise-ng-demo/modal-dialog
2. Click Example button (Opens a modal)
3. Click Cancel / Submit (Closes the modal)
4. Generate a Heap Snapshot using Chrome DevTools' Memory tab
5. Filter objects using "modal" keyword
**Note:** Repeat steps 2-3 several times and you will see an increasing number of ExampleModalDialogComponent instances from the Heap Snapshot

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,11 @@
 # What's New with Enterprise-NG
 
+## v7.0.0
+
+### 7.0.0 Features
+
+- `[Modal]` Added a new SohoModalModule that would work almost similar to SohoModalDialogModule but fixes memory leak issues. `NBCP` ([#639](https://github.com/infor-design/enterprise-ng/pull/639))
+
 ## v6.1.0
 
 ### 6.1.0 Breaking Changes
--- a/projects/ids-enterprise-ng/src/lib/index.ts
+++ b/projects/ids-enterprise-ng/src/lib/index.ts
@@ -44,6 +44,7 @@ export * from './mask/index';
 export * from './masthead/index';
 export * from './menu-button/index';
 export * from './message/index';
+export * from './modal/index';
 export * from './modal-dialog/index';
 export * from './notification/index';
 export * from './pager/index';
--- a/projects/ids-enterprise-ng/src/lib/modal-dialog/soho-modal-dialog.d.ts
+++ b/projects/ids-enterprise-ng/src/lib/modal-dialog/soho-modal-dialog.d.ts
@@ -150,6 +150,11 @@ interface SohoModalStatic {
   resize();
 
   /**
+   * Open the modal dialog.
+   */
+  open(): void;
+
+  /**
    * Close the modal dialog.
    *
    * @param destroy - destroy the html elements.
--- a/projects/ids-enterprise-ng/src/lib/modal/README.md
+++ b/projects/ids-enterprise-ng/src/lib/modal/README.md
@@ -1,0 +1,214 @@
+# SoHoXi Angular Component : Modal
+
+## Description
+
+This component provides access from Angular to the SoHoXi `modal` JQuery control.
+
+Modal dialogs allow users to enter a few key pieces of information without moving to a different page or task.
+
+### Usage
+
+Creating modal dialogs requires the injection of the service `SohoModalDialogService` into the hosting component.
+To access this service, you will need to inject `SohoModalDialogService` into the constructor of the relevant class.
+If this is a class managed by Angular (such as a Component) then adding the following will work:
+
+```typescript
+constructor(private modalService: SohoModalDialogService) {
+}
+```
+
+Angular requires a placeholder component to parent the
+dialog component when it is instantiated.  The location of the component
+is up to the calling application, but in this example the hosting component
+is used.
+
+```typescript
+@ViewChild('dialogPlaceholder', { read: ViewContainerRef }) placeholder: ViewContainerRef;
+```
+
+In the markup for the hosting component add:
+
+```html
+<!-- div #dialogPlaceholder will contain the child component -->
+<div #dialogPlaceholder></div>
+```
+
+To open the dialog, the `modal` is called on the `SohoModalDialogService`, as follows:
+
+```typescript
+this.dialog = this.modalService
+  .modal(ExampleModalDialogComponent, this.placeholder)
+  .open();
+```
+
+This returns a typed implementation of `SohoModalDialogRef`.
+
+**NOTE:**  Angular needs to be able to find the dynamic dialog component for it to be able to
+instantiate it, so you **MUST** add the component into the *entryComponents* of the hosting
+module. For example:
+
+```typescript
+@NgModule({
+  declarations: [ ExampleModalDialogComponent ],
+  exports: [],
+  imports: [ ..., SohoComonentsModule ],
+  providers: [],
+  entryComponents:[ ExampleModalDialogComponent ]
+  ]
+})
+export class ExampleModuleDialogModule {}
+```
+
+## Methods
+
+| Name | Description |
+| --- | --- |
+| `modal` | Creates a modal dialog, under the placeholder, based on the given component. |
+| `message` | Creates a modal dialog using the html (or jQuery) content |
+
+## SohoModalDialogRef
+
+| Name | Description |
+| --- | --- |
+| `open()` | Opens the dialog. |
+| `apply((c: T) => void)` | Executes the given function, with the instantiated component instance as an argument.  This is allow the concrete component to be interacted with. |
+| `close(boolean)` | Closes the modal dialog, if open.  The dialog is not closed fully until the 'afterClosed' event is fired. |
+| `options(SohoModalOptions)` | Applies the given Soho options to the modal, overrides all options. |
+| `frameHeight(number)` | Sets the extra frame height for the dialog. |
+| `title(string)` | Sets the title for the dialog. |
+| `buttons(SohoModalButton[])` | Sets the button to use for the modal dialog. |
+| `id(string)` | Sets the element id for the modal dialog. |
+| `trigger(SohoModalTriggerType)` | Sets the trigger for the modal dialog. |
+| `isAlert(boolean)` | Sets the isAlert option - controls the styling and assessibility. |
+| `content(string | JQuery)` | Defines the content of the dialog, if not using an Angular component. |
+| `cssClass(string)` | Extra CSS for the markup . |
+| `autoFocus(boolean)` | Auto Focus the dialog. TBC |
+| `beforeOpen(() => boolean)` | 'Registers a 'beforeOpen' callback, which can veto open. |
+| `afterOpen(Function)` | 'Registers an 'afterOpen' callback. |
+| `opened(Function)` | 'Registers an 'opened' callback - before the dialog is opend. |
+| `beforeClose((dialogRef: SohoModalDialogRef<T>) => boolean)` | 'Registers a 'beforeClose' callback, which can veto the close. |
+| `afterClose(Function)` | 'Registers an 'afterClose' callback. |
+| `closed(Function)` | 'Registers a 'closed' callback - before dialog is closed. |
+| `beforeDestroy(() => boolean))` | 'Registers a 'beforeDestroy' guard - vetoing the destruction. USE WITH CARE! |
+
+## Events
+
+| Name | Description |
+| --- | --- |
+
+## Examples
+
+### Simple Modal dialog
+
+This example show how a simple modal dialog component can be instantiated.
+
+```typescript
+  this.dialog = this.modalService
+    .modal<ModalDialogComponent>(ModalDialogComponent, this.placeholder)
+    .buttons([{text: 'OK', click: () => { this.dialog.close(); }, isDefault: true}])
+    .title('My Dialog')
+    .open());
+```
+
+### Simple Message dialog
+
+```typescript
+openMessage() {
+  this.dialog = this.modalService
+    .message('<span class="longer-message">Are you sure you want to delete this page?</span>')
+    .buttons([{text: 'Submit', click: () => { this.dialog.close(); }, isDefault: true}])
+    .title('Confirmation')
+    .open();
+}
+```
+
+### Setting the dialog result
+
+Setting the `dialogResult` can be achieved in several ways, the simpliest is to
+use the `close` method on `SohoModalDialogRef` in your button handler, which takes the
+dialog result as an argument.  This is set on the dialog reference and then passed into
+the `afterClose` event handler.
+
+Alternatively, you can call the `setDialogResult` method on the dialog reference, which
+also takes a dialog result as an argument.  This allow the closing of the dialog to be
+decoupled from the result.
+
+The `dialogResult` could be the _model_ used by the underlying component, or a simple status
+indicator, such as `OK` or `CANCELLED`.
+
+Here is an example for a simple message returning a status as a result:
+
+```typescript
+openMessage() {
+  const dialog = this.modalService
+    .message('<span class="longer-message">Are you sure you want to delete this page?</span>')
+    .buttons(
+      [{ text: 'No', click: () => { dialog.close('I pressed YES'); } },
+       { text: 'Yes', click: () => { dialog.close('I pressed NO'); }, isDefault: true }])
+    .title(this.title)
+    .open()
+    .afterClose(result => {
+      console.log(`The result was ${result}`);
+  });
+}
+```
+
+To provide access to any model present on the underlying dialog component, a reference to
+this component is passed to the `beforeClose`, `closed` and `afterClose` callbacks.  This
+reference can then be used to interogate the public properties of that `dialogComponent`.
+
+```typescript
+dialogRef.afterClose((result: any, ref: SohoModalDialogRef<ExampleModalDialogComponent>, dialogComponent: ExampleModalDialogComponent) => {
+  console.log(dialogComponent.model.someProperty);
+});
+```
+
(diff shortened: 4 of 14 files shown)


projects/ids-enterprise-ng/src/lib/modal-dialog/soho-modal-dialog.ref.ts
```diff
@@ -426,6 +428,10 @@ export class SohoModalDialogRef<T> {
    * @return true if the dialog can be closed; otherwise false if veoted.
    */
   private onBeforeClose(event: any): boolean {
+    this.ngZone.runOutsideAngular(() => {
+      this.componentContainer.appendChild(this.componentRef.location.nativeElement);
+    });
```
nbcp: Surprisingly, if you try to move this just below the `appendChild()` change in `onBeforeOpen()` method above, it will not leak ExampleModalDialogComponent. But of course you won't see the component inside the modal. 

I guess angular component's references won't be broken as long as you it didn't undergo its lifecycle while its outside it's original place in DOM.

bthharper: The issue here (at least when I implemented this) was that the enterprise code reparented the dialog which meant angular lost track of it.  It should be the case that calling close destroyed all the markup - looks like there are still issues with this.

bthharper: Would it be better to get the standard enterprise code to keep the dialog in the defined location?

tmcconechy: Aggree! Can we put in some sort of an issue breaking this down in EP?

nbcp: Update: I was able to get around moving the component in the DOM by doing `AppRef.attachView(childComponentRef.hostView);` because that method won't attach the component to a parent container so doing an appendChild to the modal is okay. 

However the next thing that's blocking me is when I call the method to destroy the component and modal inside the button click callback, the component still doesn't get GC'd. But if I call the method that destroys the component outside that button click callback, it's being GC'd properly..

bthharper: In the click handler try wrapping the code in angular zone: 

```typescript
this.ngZone.run(() => { ... });
```

nbcp: unfortunately still no luck with that

nbcp: yup I tested that area too a lot of times actually, so I ruled out that one out since the only remaining issue is with ExampleModalDialogComponent references.

nbcp:
I think got it fixed now 🙌! Well all except the recent full size dialog demo. But I think that one got to be fixed from enterprise and not here.
image
Recording here:
fixed-modal-memleaks

The only drawback is we cannot use vetoable events since that was one of the leaks' causes which unfortunately has no work around.

src/app/modal-dialog/nested-modal-dialog.component.ts
```diff
@@ -44,12 +40,12 @@ export class NestedModalDialogComponent {
...
-      .modal(NestedModalDialogComponent, this.placeholder)
+      .modal(NestedModalDialogComponent)
       .title(`Modal Dialog no. '${++NestedModalDialogComponent.s_depth}'.`)
       .buttons(
         [{text: 'Cancel', click: (e, modal) => { modal.close(true); }},
          {text: 'OK', click: (e, modal) => { modal.close(true); }, isDefault: true}])
-      .afterClose((f) => { NestedModalDialogComponent.s_depth--; })
-      .open();
+      .open()
+      .afterClose((f) => { NestedModalDialogComponent.s_depth--; });
```
nbcp: afterClose cannot be called before open

projects/ids-enterprise-ng/src/lib/modal-dialog/soho-modal-dialog.ref.ts
```diff
@@ -247,70 +236,91 @@ export class SohoModalDialogRef<T> {
...
+      return this.modal.element;
     });
 
+    // Create Event Observables
+    this.beforeOpen$ = fromEvent(this.jQueryElement, 'beforeopen');
+    this.open$ = fromEvent(this.jQueryElement, 'open');
+    this.afterOpen$ = fromEvent(this.jQueryElement, 'afteropen');
+    this.beforeClose$ = fromEvent(this.jQueryElement, 'beforeclose');
+    this.close$ = fromEvent(this.jQueryElement, 'close');
+    this.afterClose$ = fromEvent(this.jQueryElement, 'afterclose').pipe(
```
nbcp: vetoable events are removed but I have prepared listeners for these events in case they are needed except beforedestroy since it is almost the same with afterclose plus there's also an onDestroy hook by angular so use-case-wise it doesn't make it worth to keep.

projects/ids-enterprise-ng/src/lib/modal-dialog/soho-modal-dialog.ref.ts
```diff
@@ -247,70 +236,91 @@ export class SohoModalDialogRef<T> {
...
+      });
+    }
+
+    if (this.modal) {
+      // destroy modal
+      this.ngZone.runOutsideAngular(() => {
+        this.modal.destroy();
+        this.modal = null;
+      });
+    }
```
nbcp: this order ensures component and modal are both destroyed properly.

fitzorama:
These are breaking changes. 

Landmark uses the beforeOpen() and beforeClose() to provide functionality on a dialog. In the case of beforeOpen(), we hide a busy indicator, and for beforeClose(), we catch if an escape key was used to determine whether to allow the dialog to close if changes were made.

I am not willing to approve these changes until or unless there is a suitable replacement as this functionality was requested by our clients.

nbcp:
Yes I must admit these are breaking specially for totally getting rid of SohoModalDialogVetoableEventGuard. That one causes a lot of problems affecting other modal instances where it shouldn't be vetoing.
veto-bug

----------

I've just pushed support for vetoing on beforeclose on my most recent commit. 
vetoable-beforeclose

As for beforeOpen, would it still be necessary to have a vetoing functionality for that? I mean to me it sounds like calling open() but then prevent the modal to proceed to open. Correct me if I'm wrong on this, but you could simply not call open() at all right? 

But if we need to execute statements on those phases, I stadardized it as even hook callbacks in SohoModalDialogRef. These methods can be used as event hooks:
-   beforeOpen()
-   opened()
-   afterOpen()
-   beforeClose() <- can prevent closing
-   closed()
-   afterClose()

tmcconechy:
The code is in place in EP such that any of the beforeXXX events/callbacks you can return false to not open or close. So yes it is valid and there have been use cases where its used.

Although beforeClose is more common. So we probably just need to be able to veto the before XXX events.

projects/ids-enterprise-ng/src/lib/modal-dialog/soho-modal-dialog.service.ts
```diff
@@ -43,29 +41,12 @@ export class SohoModalDialogService {
    * The dialog won't necessarily be open yet, see the dialog ref api for further methods.
    *
    * @param component - the type of the component to instantiate; must not be null.
-   * @param parent - the parent container; must not be null.
-   * @param options - the control options to use.
    *
    * @return the modal dialog reference.
    */
-  public modal<T>(component: ComponentType<T>, parent: ViewContainerRef, options?: SohoModalOptions): SohoModalDialogRef<T> {
+  public modal<T>(component: ComponentType<T>): SohoModalDialogRef<T> {
```
fitzorama: I suggest leaving the parent argument in place and making it optional and marking it deprecated for a period of time as this is also a big breaking change.

fitzorama:
@nbcp A couple of ideas, one which I indicated earlier off-line.
- Adding a brand new soho modal dialog that incorporates your changes. This moves the risk to the new component and we can eventually deprecate and remove soho-modal-dialog.
- Instead of trying to get 'all' of the memory leak fixed, can we could fix as much as we can without introducing the breaking changes? I'm thinking of where the order mattered on destroy, etc.

nbcp:
I think adding the beforeXXX vetoable callbacks would be alright.

Hmm would it be better to move this as a separate module then? Maintaining everything from the old one and let those who uses this library decide when to switch over to the new one? I think that maybe the fastest way to get this out. Since this also breaks SohoMessageRef that also relies on SohoModalDialogVetoableEventGuard.

I am also not a fan of passing callbacks and chaining methods and boilerplates `return this;` so by moving this into separate module I can replace those with observables and clean this up to get a much simpler readable code.

projects/ids-enterprise-ng/src/lib/modal-dialog/soho-modal-dialog.service.ts
nbcp: yes this was because this now uses ApplicationRef instead of ViewContainerRef so that the passed component won't need to be displayed temporarily some placeholder somewhere before being  moved inside the modal content.

bthharper: I agree with @fitzorama on this, we need to keep the existing APIs until we increment the major version.

tmcconechy: That could line up with Ng 9 version 7/0 NG?

src/app/modal-dialog/nested-modal-dialog.component.ts
bthharper: This would be subtle breaking change - has something changed in how this is implemented, previously I had an `Subject` to handle that.

projects/ids-enterprise-ng/src/lib/modal-dialog/soho-modal-dialog.ref.ts
bthharper: @tmcconechy  - did we check if EP 'moving' the dialog's markup was causing the issue here?

tmcconechy: @bthharper I did not check that? Which issue is it?

projects/ids-enterprise-ng/src/lib/modal-dialog/soho-modal-dialog.service.ts
pwpatton: The reason we used a ViewContainerRef (placeholder) was so that "nested" dialogs would work. If we attach a dialog to the ApplicationRef and then that dialog opens another dialog does attaching that to the ApplicationRef work?

bthharper:
I guess it's a preference, but we use fluent APIs quite alot in our applications, but that might just be a .NET thing.  

I think it is important that we keep the APIs semantically and syntactically consistent until be bump the major version anyway.

As I mentioned earlier, when I wrote this originally code there were issues with the fact that EP moved (copied) the dialog markup when you opened the dialog, have we investigated that?

I wonder if angular is losing track of the component and not able to tidy up.



.

projects/ids-enterprise-ng/src/lib/modal-dialog/soho-modal-dialog.ref.ts
bthharper: I remember (it was a while ago) that when the dialog was displayed (opened) EP would move (or copy) the markup to a placeholder before displaying it.   This did cause issues destroying the markup which we had to fix.  I missed the memory leaks, but I was wondering if angular was not getting told to remove the component - and any references?

tmcconechy:
Would this be the issue @bthharper  https://github.com/infor-design/enterprise-ng/issues/135 ?

tmcconechy:
Yes it does move it to a container that can be centered on the body

nbcp:
I feel like we are going in circles, similar to the references in these modal objects here 😆 My apologies if I haven't been very clear in explaining everything about the problems and solutions. I believe this is a very critical problem to solve so I'll try my best to summarize everything I know & remember so far:

### MAIN PROBLEM: "Memory Leaks in Soho Modals"
image

_Originally posted by @nbcp in https://github.com/infor-design/enterprise-ng/issues/636#issuecomment-545353428_

For every modal opened, the component's instance (modal's body), SohoModalDialogRef, and Modal Objects are not being freed from memory. So as you open/close more modals, it just consumes more unnecessary objects in memory.

### CAUSES OF PROBLEM:

1. **SohoModalDialogVetoableEventGuard** - this is an interface you can implement on a component where you can define the vetoable event callback functions (beforeOpen, beforeClose, beforeDestroy), which when returns false prevents opening/closing of the modal. The problem with this design is we are holding a reference to the component's instance `this.eventGuard = componentRef.instance;` even if we're not using those vetoable events. 
2. **SohoModalDialogInjector** - this replaces the injector to be passed into the component so that it would also include SohoModalDialogRef to be accessible inside the component. So the problem here is SohoModalDialogRef is dependent on the component but the at the same time the component is also dependent to the SohoModalDialogRef, keeping both objects from getting destroyed.
3. **Unclosed Observables in SohoModalDialogRef** - open$, close$, afterOpen$ subscriptions in SohoModalDialogRef are not closed properly resulting to not letting go of their references
4. **Returning component dialog instance on close and afterClose event callbacks** - after these events, the modal and component instances needed to be destroyed. However passing the component instance back to the calling function, then from there having something to hold the component's reference would prevent from getting destroyed.
5. **Moving component DOM after parent.createComponent()** - parent is a ViewContainer, basically a placeholder in another component. We are creating and attaching the component in that placeholder and then later move its DOM to be used inside the modal's body. Not sure of the exact details but Angular's lifecyle respects the hierarchy of components so parent-child components have references with each other. Moving the component away from its original DOM would break some of those reference ultimately preventing its object from getting destroyed.

### SOLUTION:
1. We can remove SohoModalDialogVetoableEventGuard since beforeOpen, beforeClose, beforeDestroy methods don't necessarily need to reside in the component to be placed inside the modal. You may use these methods from SohoModalDialogRef.
2. We can remove SohoModalDialogInjector since SohoModalDialogRef is already returned by SohoModalDialogService.modal() method, you may already use that reference to access it methods. Plus it would make the component more flexible (not needing that extrainjector) in cases where you also use it outside of the modal.  
3. Trivial fix, just closed these observables on destroy.
4. A bit related to `#3`. Since on these events, we'll start destroying the modal, it would be best to not allow outside force to hold a reference to the component. We need to destroy the component before we destroy the modal and return the result on afterClose. This would mean that even if we return the component's instance at that point, it will just be null. So its better to not return the component instance at all on these events.
5. Using ApplicationRef.attachView() would create the component's instance and make Angular change detection work but not attach it to the DOM yet. This would give us freedom to place the component somewhere else at a later time like the modal's body. This also keeps us free from stale DOM reference Angular-wise. This also frees us the responsibility of having to define a placeholder for each modal we open since it's just a temporary place to hold the component before the modal is created. And yes, nested/stacking modal still works perfectly.

nbcp:
my last update on this branch is to separate all my changes into a separate module and retain the old one

bthharper:
The changes look fine to me, and thanks @nbcp for the clarification - that really helped.   My main concern was the breaking changes in the Api, but we can give notice now on those.

nbcp:
Updated the CHANGELOG and README for the API changes. Also fixed some issues found in afterclose and veto event methods and updated the demo to show how those works.
modal-veto

bthharper:
@nbcp I wonder if any of the memory leak fixes could be pushed into the original version (for this release).  I was thinking specifically of no. 3 and 5.  Just a thought.

bthharper:
> @nbcp @pwpatton @bthharper - was thinking if we push this then we just make next version 7.0.0 but of course have to wait for NG 9 and the work needed on that. What do you think?

Angular 9 is due soon (November), but I've not done any testing so far - not expecting it to be an issue although enabling ivy is more problematic.

Personally, I would include these changes in the current release as they are non breaking, and just the addition of a new feature.  It would give everyone some time to migrate.

tmcconechy:
Im approving this fix - conditionally. But we will merge this into master after 6.x is created in a week or two. So holding this off for right now

nbcp:
> @nbcp I wonder if any of the memory leak fixes could be pushed into the original version (for this release). I was thinking specifically of no. 3 and 5. Just a thought.

@bthharper, those 5 solutions is more of an all or nothing fix. These object references are entangled with each other. Picking just 2 solutions while leaving the rest would not really free us from any of the leaking objects.

bthharper:
I had a little look at this (whilst fixing the message class on the dialogs) and think your changes are awesome.  It's made me revisit our applications to ensure we are tidying up any subscriptions, I'd read about about using `takeUntil` before and it's a neat fix.

I did manage to port the changes to the original code (keeping the same api), and now have no leaks.  However, I have hit the same issue you mentioned on the `vetoable` interface.   I'll spend a bit longer on that, maybe we can include it in your upgraded version, but may well come to the same conclusion as you that it would be easier to simply mark as obsolete and remove.

fitzorama:
@tmcconechy No objections

tmcconechy:
based on @bthharper comments he was looking at this so just waiting for a nod from him

PR #663, opened 2019-11-15
Merged Modal Memory Leak Fixes to existing component.

bthharper:

To avoid breaking API changes, I have merged the fixes from the new modal code back to the original service, keeping the API as previously defined.  

The decision whether to move over to the new component can then taken when the major version number is bumped.

#636 

1.  Go to http://localhost:4200/ids-enterprise-ng-demo/modal-dialog
2. Click Example button (Opens a modal)
3. Click Cancel / Submit (Closes the modal)
4. Generate a Heap Snapshot using Chrome DevTools' Memory tab
5. Filter objects using "modal" keyword

**Note:** Repeat steps 2-3 several times and you will see an increasing number of *ExampleModalDialogComponent* instances from the Heap Snapshot


docs/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - `[Modal]` Added a new SohoModalModule that would work almost similar to SohoModalDialogModule but fixes memory leak issues. `NBCP` ([#639](https://github.com/infor-design/enterprise-ng/pull/639))
 - `[Datagrid]` Made the last two options on exportToExcel optional. Note that passing in a third option is incorrect and will give you no headers in the exported file, so generally shouldnt be used. `TJM` ([#654](https://github.com/infor-design/enterprise-ng/pull/654))
 - `[Dropdown]` Added keydown output EventEmitter. `NBCP` ([#658](https://github.com/infor-design/enterprise-ng/pull/658))
+- `[Modal]` Merged SohoModalModule into SohoModalDialogModule keeping current API but fixing the memory leaks. `BTHH` ([#663](https://github.com/infor-design/enterprise-ng/pull/663))
```
tmcconechy: Can you add a section for 7.0 and put this here. This branch is moving towards 7.0
If we need this fixed in 6.3.0 i can make a branch for that. Presently that would be off 6.2.x 

Or does this negate the need for a 7.x?

pwpatton:
We'd like to verify this branch change against landmark before we promote it. Going to make as work in progress so it doesn't accidentally get promoted.

bthharper:
Totally agree waiting for verification, I wanted to push these changes up to start the discussion on whether we want to include these pre 7?  Given Angular 9 is not out yet, and it can take a while to get the upgrade done perhaps we need a new point release of ids-enterprise-ng?

I also have changes for the context-action-panel, which has the same memory leaks, do we want to include those here or in a separate PR?

tmcconechy:
@bthharper if its breaking lets keep it on master - which will be 7.0 once NG 9 is released and done.
We will almost 100% sure do a 6.3.x this will be off the 6.2.x existing branch and i can do that once 6.20 is release this week.

If the change is minor you can just put it on the 6.2.x for 6.2 or wait for that branch which i can only do later this week.

bthharper:
@tmcconechy  - no problem.  

The intention was to make this a non braking change to the existing modal API (and context action panel) -  it seems to work okay in the demo samples, but pushing out this week would definitely be risky.  I'll try a local drop of the components against our apps, maybe @fitzorama can do the same?  If it looks okay maybe we can push into 6.3.

Not totally sure how soon Angular 9 will be out, its RC currently but it could be several more weeks, plus time to implement any local changes required.  It took us 2-3 months last time.

tmcconechy:
Yeah took a few months last time. But was hoping to shave that down. We did make a decision to wait longer. As the branch was ready for a while.

bthharper:
@tmcconechy Could be merge this then create a dev release we can test, I'm not seeing the same issue as @pwpatton.

What do you think?  We can revert afterwards?

tmcconechy:
Thats fine with me, but @pwpatton is holding changes for @fitzorama to test. I think its ok since we have a 6.3.x dev branch for next release for NG. Master is 7.0 leaning for later now.

fitzorama:
We have resolved the lm-locale-initializer.service error we were previously receiving. In testing this change, I am seeing SohoModal* named objects in the Heap Snapshots that do not seem to get destroyed. I am not sure I am reading the Memory tab correctly. Screen shot below displays the Memory tab with three snapshots: 
- Dialog opened and closed, Snapshot 1 taken
- Dialog opened and closed twice more, Snapshot 2 taken
- Dialog opened and closed once more, Snapshot 3 taken
image

My question is are the SohoModal* items to be destroyed as well?

bthharper:
@fitzorama - is this running against your app?  In the demo I see the following:

image

This is after opening and closing all the modal dialog twice on the demo page.

fitzorama:
@bthharper Yes, running against the Landmark application and similar results.

fitzorama:
@bthharper So just the demo modal components are being destroyed and the SohoModal components are not? It feels like there is still a memory leak. Is there something that the application needs to do to or is this as good as it is going to be for now?

@nbcp any thoughts?

docs/CHANGELOG.md
pwpatton: I think until we have an `ng9` PR ready we should hold off on a `7.x`. @bthharper where are you on that one Theo?

bthharper:
@fitzorama I believe this is correct, as you are seeing the component definitions included in the owning module, which are there to make sure the component is not excluded from the final build by the transpiler and tree shaker.

docs/CHANGELOG.md
tmcconechy: Sure it can be whenever but the master branch has a breaking change in it that will require a 7.x release? Unless that was fixed?

bthharper: @pwpatton  - not that far down the road, was waiting for the release to look at it.  I do have an IVY branch that I was working on.    

@tmcconechy - what's the breaking change?

tmcconechy: I thought the previous https://github.com/infor-design/enterprise-ng/pull/639 PR. But if we are happy to release that into a 6.3 maybe we are OK?

tmcconechy: Should we just merge everything in master and make it a 6.3 next even though there is a minor breaking change?

tmcconechy: on discussing with Phillip lets keep master going to 6.3.x for now. Despite that change. 
So all this can go in master.

nbcp:
yeah.. this pr is not a leak-proof fix, since this more of a sub that is leaning more towards non-breaking changes, it just reduces the overall leaks. The eventGuard holding reference to the component's instance is is one of the reasons causing the leak. So if we really want the leak-proof fix which is already in master, the real question would be: Should we wait for Angular 9 with before we release version 7? Or should we just release 7 and have Angular 9 for version 8?

tmcconechy:
I guess we could just call the next version 7.0 just because of the one modal change. Since we want to stick to semantic versioning.

But where are the notes on the change again? Not seeing anything obvious in the log in terms of what people need to do?

https://github.com/infor-design/enterprise-ng/blob/master/docs/CHANGELOG.md#620-fixes

pwpatton:
@tmcconechy This PR is not a breaking change. And the other PR that @nbcp did isn't either as it's a new module with a new modal wrapper and doesn't require that people change to it.

tmcconechy:
Ok i was really confused about that all. Thanks for clarifying. Lets continue with 6.3.x for next version

bthharper:
@pwpatton @fitzorama - ready to commit on this one?

PR #678, opened 2019-12-04
New code block implementation example

### LINKED ISSUE #675: New code block implementation
Add an example for a new lm code block implementation.

We would like how code block works to change. Instead of having flexing display only to input fields, on the form the code block should display as read only. If the field is clicked on, or tab'ed into, a popover should display with editable fields. Then these fields could share an ngModel with what is displayed on the form, so changes are propagated back to the form.

When the popover opens, the first focusable field should be focused. The editable fields should be able to be tab'ed thru, and when the last field is blurred out of, the next field on the form should be focused.

If the escape key is pressed while in the editable fields popover, the popover should close and the next field on the form should be focused.

@tmcconechy is there someway to set the width of a tooltip that I am missing?

Previous implementation was difficult to get correct. This is hopefully a cleaner implementation, that power users can use without multiple tabs per editable field.

**Additional context**
image

NewCodeBlockExample

Sovia:
Adding a ng example for a new code block implmentation. This was to help prototype the new implementation, and to help provide an example for any issues found.

Closes #675 

Pull in the changes, and go to the code block component. See that it matches how it should work (described in 675)

DifferentFieldsCodeBlockPopover

There is still an issue with tabing, but wanted to get this example merged in before debugging.

### DIFF
--- a/src/app/code-block/code-block.demo.html
+++ b/src/app/code-block/code-block.demo.html
@@ -1,202 +1,98 @@
 <div class="row top-padding">
-
   <div class="twelve columns">
+    <h2>Using the soho-popdown to display the editable code block fields.</h2>
+    <p>
+      1. Can click into the code block field, and the popup opens up, and the first editable field is focused on<br>
+      2. Can tab into the code block field, and the popup opens up, and the first editable field is focused on<br>
+      3. Tabbing thru the editable fields, the popdown will close when the last focusable element is blurred<br>
+      4. Pressing esc key will close the popdown where ever. The next field on the form gets focused<br>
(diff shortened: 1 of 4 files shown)


tmcconechy:
Maybe this is fine to merge for now but seeing some minor issues..

- the popup is "flat" it should have some wondering if it should have some shadows like http://master-enterprise.demo.design.infor.com/components/popover/example-index.html
- im not sure how you close the popup when its open
- needs more padding on the popup (20px around)

Sovia:
Currently you can close the popover by hitting the escape key, or blurring out of the last field in the poover.

Issue #712, opened 2020-01-22
Calendar/Week-view-Template parse errors when running "ng test" due to the "on" prefix on the input names.


vonnyw:
**Describe the bug**
Was getting error: 
Template parse errors:
Binding to event property 'onRenderMonth' is disallowed for security reasons, please use (RenderMonth)=... 
'onRenderMonth' is a directive input, make sure the directive is imported by the current module.

https://github.com/angular/angular/issues/11756

**To Reproduce**
1. Run ng test
2. Got the above error.

**Expected behavior**
No error when running ng test.

**Version**
- ids-enterprise-ng: v6.4.0

**Platform**
 - OS Version: Windows 10
 - Browser Name Chrome
 - Browser Version 79.0.3945.117 (Official Build) (64-bit)

PR #715, opened 2020-01-23
Calendar test failure

vonnyw:

Fix ng test failure

https://github.com/infor-design/enterprise-ng/issues/712

Run ng test - tests should be clean

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## v6.4.0
 
+### 6.4.0 Breaking Changes
+- `[Calendar]` - Removed "on" prefix on any of the inputs due to test error. `VW` ([#712](https://github.com/infor-design/enterprise-ng/pull/712))
+
 ### 6.4.0 Fixes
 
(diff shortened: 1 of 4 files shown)


vonnyw:
@tmcconechy Was getting error:
Template parse errors:
Binding to event property 'onRenderMonth' is disallowed for security reasons, please use (RenderMonth)=...
'onRenderMonth' is a directive input, make sure the directive is imported by the current module.

Here's more info: angular/angular#11756

tmcconechy:
Ok, was the issue only in the tests? If so i suggest we just xit the tests until the bug is fixed. That way we dont need to break the API and make things a bit less consistent.

Why is it just the calendar ? Not all with `on` ?

vonnyw:
The issue is only in tests, correct. We can wait to push this at a later time if it makes the build process a little easier. I originally made the change in calendar and week-view components, other components probably don't use the "on" prefix on inputs, outputs are okay.

tmcconechy:
Ok, i see. These are inputs. So yeah thats a bit wierd. one thing i still dont get is why the tests passed before and now fail ? Do they need updating?

vonnyw:
Not sure why the test didn't fail before, yes, they do need updating as well.

tmcconechy:
Closing this for now as may not be needed.

PR #727, opened 2020-02-05
Added IE configuration

### LINKED ISSUE #668: General: Demo app is not running in IE 11
**Describe the bug**
Out of the box the demo app does not run in IE. To solve this differential loading (great feature) must be disabled. So teams need to weigth the use of this with the need to support IE.

And we need to decide if the demo app should target `es6`. Solution is on and discussed here:

https://stackoverflow.com/questions/56563742/how-do-i-make-angular-8-compatible-with-ie11

**To Reproduce**
1. `npm run start`
2.  Open IE 11
3. Load http://localhost:4200/ids-enterprise-ng-demo/
4. See see that it just has a busy indicator and no error

**Expected behavior**
Works (optionally)

**Version**
- Latest (6.x, 7.x)

**Platform**
 - Browser Name IE 11

**Additional Thoughts / FIxes**

- [ ] Do we by default target es5 or not as described in the link? Or deprecate IE 11
- [ ] Document this in the docs how to set this.
- [ ] Add a second build command where you can launch if IE 11 is needed.

bthharper:

Add support for running the demonstration application in IE, by setting the TypeScript target to es5.

Closes #668.

```sh
git clone 
npm i
npm run build:lib
npm start:ie
```

Navigate to http://localhost:4200/ids-enterprise-ng-demo

<!--


tmcconechy:
Does this only have an impact if started in ie mode?

bthharper:
It only impacts running the demo when choosing the IE compatible flag.

PR #730, opened 2020-02-07
Upgrade to Angular 9

### LINKED ISSUE #586: Upgrade to Angular v9 and Ivy
At the moment, the users of `ids-enterprise-ng` often needs to wait for quite long time to upgrade to the next major version of Angular. This is not flexible at all.

A plan to prepare for Angular v9 with Ivy compiler. The reasons for the preparation are:
- #578 of `ids-enterprise-ng` was finished.
- Ivy was enabled as default compiler in the latest unreleased version of [Angular](https://github.com/angular/angular/blob/master/CHANGELOG.md#900-next3-2019-08-21)

I think it's wise to prepare before a major version of Angular is actually released. This will help users of `ids-enteprise-ng` to be more flexible to upgrade their apps to next major version of Angular instead of waiting for `ids-enterprise-ng` to be compatible. 

**N.A.**

**Additional context**
The reason for early preparing is assuming that `ids-enterprise-ng` finishes being compatible with next major version of Angular before the major version is released, then when the major version of Angular releases, `ids-enterprise-ng` can release at the same time. As a result, the users of `ids-enterprise-ng` can enjoy the new major version.

bthharper:

Upgrade of demo application and library to angular 9.

Closes #586.

Clone repository

```sh
npm i
npm run build:lib
npm run lint
npm run test:lib
npm run start
```

Navigate around application.

We may want to hold off merging this until the release date of 7 is agreed, are we going to keep 6.5 for example, or release 6.5 then 7 simultaneously.


projects/ids-enterprise-ng/src/lib/datagrid/soho-datagrid.spec.ts
```diff
@@ -176,7 +176,7 @@ describe('Soho DataGrid Unit Tests', () => {
   beforeEach(() => {
     TestBed.configureTestingModule({
       declarations: [SohoDataGridComponent]
-    });
+    }).compileComponents();
```
pwpatton: @bthharper is this necessary with ng9?

src/polyfills.ts
```diff
@@ -19,32 +19,32 @@
...
+// import 'core-js/es/parse-float';
+// import 'core-js/es/number';
+// import 'core-js/es/math';
+// import 'core-js/es/string';
+// import 'core-js/es/date';
+// import 'core-js/es/array';
+// import 'core-js/es/regexp';
+// import 'core-js/es/map';
+// import 'core-js/es/weak-map';
+// import 'core-js/es/set';
```
pwpatton: so what is the plan here, I think we have to at least support ie11 for a little while longer.

src/test.ts
```diff
@@ -18,11 +18,3 @@ getTestBed().initTestEnvironment(
...
 // And load the modules.
 context.keys().map(context);
-
-// HACK: Chrome XX disconnects if there are too many synchronous tests in a row
-// because it appears to lock up the thread that communicates to Karma's socket
-// This async beforeEach gets called on every spec and releases the JS thread long
-// enough for the socket to continue to communicate.
-// The downside is that it creates a minor performance penalty of around 10-15%
-// increase in the time it takes to run out unit tests.
-// beforeEach((done) => done());
```
pwpatton: Is this no longer an issue?

tsconfig.json
```diff
@@ -40,7 +40,8 @@
   },
   "angularCompilerOptions": {
     "fullTemplateTypeCheck": true,
-    "strictInjectionParameters": true
+    "strictInjectionParameters": true,
+    "enableIvy": false
```
pwpatton: so are we not enabling Ivy in ids-ep-ng 7?

projects/ids-enterprise-ng/src/lib/version/version-initializer.service.spec.ts
```diff
@@ -6,12 +6,12 @@ describe('Soho Version Initialiser Service', () => {
...
     imports: [
     ],
-     providers: [
+    providers: [
     ]
   }));
 
   it('should be created', () => {
-    const service: SohoVersionInitializerService = TestBed.get(SohoVersionInitializerService);
+    const service: SohoVersionInitializerService = TestBed.inject(SohoVersionInitializerService);
```
pwpatton: So in all tests it's inject instead of get?

pwpatton:
Are we doing to make a 6.5.x branch so fixes can be made before moving to ng9?

src/polyfills.ts
bthharper: You can still use the trick in the `tsconfig` to target es5.

src/test.ts
bthharper: Have not had any issues.

tsconfig.json
bthharper: Not for the demo app, as there is (was) an issue with SVG icons.  See the CHANGELOG for details.

projects/ids-enterprise-ng/src/lib/version/version-initializer.service.spec.ts
bthharper: Yep - get is deprecated.

tmcconechy:
yes lets make a branch just in case for 6.5x but would rather add this sooner than later. i will review this and make a branch today.

bthharper:
> @tmcconechy @bthharper Are we doing to make a 6.5.x branch so fixes can be made before moving to ng9?

I think doing a 6.5 and then a 7 is a good idea.  I guess we could release an early access angular 9 version, but no hurry.

tmcconechy:
There is one problem with the build / tests https://travis-ci.com/infor-design/enterprise-ng/jobs/284954392#L275-L279 not sure why this fails as i cannot get into the file as its on the server. 

Any guesses? I saw a reference to node ? did this change  by any chance?

```
$ node --version
v10.16.3
```

bthharper:
I added a quick fix for the Travis build, it does highlight the fact I rarely run `npm build` and with the application size and multiple targets this uses a lot of memory during compilation, almost 4GB on the mac - which also failed when I tried it locally.  

Do you think it would be better to add a specific Travis build which uses the "modified node" call?

tmcconechy:
That sounds good to me? High vs low memory mode commands? We can change which one runs in Travis.yml

tmcconechy:
OK to merge this @pwpatton @EdwardCoyle I made a 6.5.x branch for 4.27. This can go out with 4.28 and live on master. Will make a dev build for it to test out.

pwpatton:
@tmcconechy  I think so

PR #737, opened 2020-02-18
732 - Textarea initial grow

### LINKED ISSUE #732: Textarea: Autogrow does not work in Angular when content arrives
From the chat:
Any one have idea how the autogrow property on  soho-textarea works? In my case it is setting an initial height  which is irrespective of the content in the textarea , also if i edit the textarea it is auto growing again. How do i make it autogrow based on text area content on intial loading.

Tim: Maybe in angular this doesnt work as the content is there later? You could try calling updated after the content is updated or maybe we need to tweak the directive a bit

tmcconechy:

Bumps the dev version and updates the textarea example to show that it can grow on initial grow.

Fixes #732

- go to http://localhost:4200/ids-enterprise-ng-demo/textarea
- make sure your browser is big enough so the menu is open (may get an issue it doesnt size as the menu squished the page that can't really be fixed
- notice that the one field is expanded now on initial load


bthharper:
I think that's an intermittent failure, as I had the same locally once.  @pwpatton / @tmcconechy - something about calendars on destroy.

PR #745, opened 2020-02-20
Demo Application Personalize Fixes

bthharper:

The Theme menu was not working after updating the header to use the toolbar-flex component.  

N/A

```sh
npm i
npm run build:lib
npm run start
```

Navigate to http://localhost:4200/ids-enterprise-ng-demo, and change the theme and/or colour.


bthharper:
@tmcconechy - this is mostly working now, but for reason the events are not first the first time the theme / color is changed.  Would this be a problem in the EP code?

The other problem I saw was that the colors are appearing in upper case sometimes and in lower case other times which is breaking the selection code.  I could just add toLower/toUpper however, it would be better to understand what's happening.

This has not fixed my application, which is reloaded when I select any theme / color but that's not the case in the demo.

bthharper:
> Lets look at this some more. See a few weird things.
> A)
> 
> 1. start out on Soho Light - reloading if necessary
> 2. select soho dark -> check console and see an error - is it missing `data-theme-name` and `data-theme-variant` in the right places?
> 

I'm not seeing any error when I run the example, which is odd.  The initial theme load even I am getting is because the locale is being loaded by the APP_INITIALIZER ```Soho.Loclale.set(...)```.

Then I load the Theme from `localStorage` and call `setTheme().`

> B)
> 
> 1. start out on Soho Light - reloading if necessary
> 2. look for the icon in the application menu - its missing? Something with the setting the correct icon set maybe?

Which icon?  The Application Menu icon?

image

> 
> C) I have a fix for the missing text in [] it will be in tomorrows dev build

:-)

tmcconechy:
@bthharper i made some changes and me and @EdwardCoyle are looking at the error/first load issue. Seems flex toolbar related. So we will update this from here.

EdwardCoyle:
@bthharper try giving this a shot one more time after https://github.com/infor-design/enterprise/pull/3530 is merged.  I tested locally and it seems to fix the issue.

tmcconechy:
Thanks @EdwardCoyle we did the nightly already today. So will put this in tomorrows nightly. I fixed a couple other things (black text on headers hopefully). Once tomorrow mornings night its done we can put this all together and finalize

bthharper:
Thanks @EdwardCoyle / @tmcconechy , will look tomorrow once the new dev build is out.  Still having issues in our local app, but will wait for this till I investigate further.

tmcconechy:
The PR missed the daily build (my mistake). So this now wont be ready until tomorrow AM.

tmcconechy:
OK this should be all good now. Please have a look folks. I also fixed https://github.com/infor-design/enterprise/pull/3538 but that can come in next time we update after it gets merged

bthharper:
Tested successfully in my local application, no more screen refreshes and the theme changes appropriately.

bthharper:
@tmcconechy  @pwpatton  @EdwardCoyle @fitzorama  we can push this through now.

PR #773, opened 2020-04-03
#772 added feature to replace the previous button label in SohoStepProcess…

### LINKED ISSUE #772: StepProcess: Add feature to replace the previous button label in SohoStepProcessComponent
SohoStepProcessComponent currently only supports replacing the next button label but not the previous button label

Add another class field for the previous button.
`@Input() set previousButtonLabel(label: string) {`
`  if (this.jQueryElement) {`
`    this.jQueryElement.find('.js-step-link-prev').html(label);`
`   }`
`}`

JBismonte:
…Component

added a feature to change previous button label. rather than just the next butotn.

Closes #772 

no issues when  the following test ran:  "npm run testdebug"

### DIFF
--- a/projects/ids-enterprise-ng/src/lib/stepprocess/soho-stepprocess.component.ts
+++ b/projects/ids-enterprise-ng/src/lib/stepprocess/soho-stepprocess.component.ts
@@ -169,6 +169,12 @@ export class SohoStepProcessComponent implements AfterViewInit, OnDestroy {
       this.jQueryElement.find('.js-step-link-next').html(label);
     }
   }
+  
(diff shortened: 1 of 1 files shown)


tmcconechy:
I will force merge this i fixed the build problem in master. should be ok

nbcp:
I don't like this solution. JQuery's html method is vulnerable to XSS attacks. I'm confused why do we need to do this instead of using angular's template syntax to set the previous and next button strings.

pwpatton:
Ben. Agreed and easy enough to change w/o changing api of the component.

bthharper:
Do we need a new ticket?

PR #820, opened 2020-05-18
[soho-field-filter] A more convenient way to set the filter operator

nbcp:
Currently it is not very convenient to programmatically change the selected operator in the filter field. Either you update fieldSettings or fieldDropdownDataSet Input which would consist of the whole array of SohoFieldFilterOption. Never had luck in using setFilterType via ViewChild in the parent component, this is very also inconvenient esp. if you have a lot of soho-field-filter in the template.

This new change adds a new Input named **selectedFilterType** which allows for a more convenient way to set the filter field operator. Also added a new type named **SohoFieldFilterOperator** to have stricter type checking for the input.

NA

1. checkout this branch
2. go to http://localhost:4200/ids-enterprise-ng-demo/field-filter
3. switch values from the "Filter Operator for Text Field" dropdown
4. see realtime changes in the "Text Field" operator

### DIFF
--- a/projects/ids-enterprise-ng/src/lib/field-filter/soho-field-filter.d.ts
+++ b/projects/ids-enterprise-ng/src/lib/field-filter/soho-field-filter.d.ts
@@ -44,18 +44,21 @@ interface JQuery<TElement = HTMLElement> extends Iterable<TElement> {
   fieldfilter(settings?: SohoFieldFilterSettings): JQuery;
 }
 
+type SohoFieldFilterOperator =
+  'end-with' | 'does-not-end-with' |
+ 'start-with' | 'does-not-start-with' |
+ 'equals' | 'does-not-equal' |
+ 'contains' | 'does-not-contain' |
+ 'calendar' | 'in-range' |
+ 'is-empty' | 'is-not-empty' |
+ 'less-equals' | 'less-than' |
+ 'greater-equals' | 'greater-than' |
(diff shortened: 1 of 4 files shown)


projects/ids-enterprise-ng/src/lib/field-filter/soho-field-filter.d.ts
```diff
@@ -44,18 +44,21 @@ interface JQuery<TElement = HTMLElement> extends Iterable<TElement> {
...
+ 'start-with' | 'does-not-start-with' |
+ 'equals' | 'does-not-equal' |
+ 'contains' | 'does-not-contain' |
+ 'calendar' | 'in-range' |
+ 'is-empty' | 'is-not-empty' |
+ 'less-equals' | 'less-than' |
+ 'greater-equals' | 'greater-than' |
+ 'between' | 'selected-notselected' |
+ 'selected' | 'not-selected' |
+ 'sort-a-to-z' | 'sort-z-to-a';
```
nbcp, line 58: Btw just found out, this looks identical to SohoDataGridColumnFilterConditions in soho-datagrid.d.ts. However that one lacks some of the values, I wonder which one is more correct?
Here's their difference: (left: SohoFieldFilterOperator, right: SohoDataGridColumnFilterConditions )
image

nbcp:
realize it won't work during OnInit sine the jquery element will only be built AfterViewInit so I've updated the fix to take that into account

PR #822, opened 2020-05-20
[soho-field-filter] field filter reset

nbcp:
Forgot to consider reset in #820. You may set selectedFilterType to any default value that you want you perform the reset, but normally you would expect just set operator and value to null and let the directive do its job to reset to the default.

improvement from #820

1. checkout this branch
2. go to http://localhost:4200/ids-enterprise-ng-demo/field-filter
3. change "Text Field" operator / value
4. hit the reset button
5. value and operator must reset 

field-filter-reset


nbcp:
This maybe convenient but it will inevitably cause **ExpressionChangedAfterItHasBeenCheckedError** since the directive will emit a filtered event that when used to update the model, Angular would detect that the null has become something else e.g. 'equals' or 'contains'. So closing this now as this method wouldn't work when running angular in production

PR #867, opened 2020-07-13
700 - Implement NG Component Wrappers for Breadcrumb

### LINKED ISSUE #700: NG: Breadcrumb component
The breadcrumb UI pattern is widely used and effectively gives the user an overview of where she is in a given navigation path. It also provides a simple way to navigate to previous states or pages. Today there are css classes in IDS that can be used to create breadcrumbs, but there is no component with clear examples and the current breadcrumb implementation is lacking two important features.
- Correct truncation.
- Possibility to see truncated parts of the breadcrumb path 

An extension of what exists in _\src\app\datagrid\datagrid-toolbar.demo.ts_
1. A component such as
`<soho-breadcrumbs>`
2. Input similar to datagrid-toolbar.demo.ts
3. Output which fires when a breadcrumb is clicked.
4. Breadcrumbs should be truncated when they don't fit on the screen and the lowest level breadcrumb should always be visible. Today it's the first thing that disappears.
5. It should be possible to view and click parts of the breadcrumbs that has been truncated, so the user can navigate to that state. One example of this is the file system navigation bar on most computer operating systems.

Today we are using the existing classes to build breadcrumbs, like the example in \src\app\datagrid\datagrid-toolbar.demo.html

This implementation lacks what is stated above so each developer must create their own solution (including how breadcrumbs are being selected). Also the lack of an actual component makes it less probable that developers use the IDS solution, making the look inconsistent between apps.

EdwardCoyle:
This PR adds the following:
- an Angular component wrapper for the IDS Enterprise Breadcrumb component
- Typescript definitions for the IDS Breadcrumb
- "Gauntlet" demo page for testing the API 
- Unit tests

Closes #700 

- Pull this branch, build, run the NG demoapp
- Open http://localhost:4200/ids-enterprise-ng-demo/breadcrumb - this is a simple demo showing a standard breadcrumb setup.
- Open http://localhost:4200/ids-enterprise-ng-demo/breadcrumb-gauntlet - this is a more complex "gauntlet"/sandbox page using the API to add/remove breadcrumbs from the list.  Try all the controls and ensure everything is working correctly.
- Ensure tests pass

----
@gushanson

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### 7.3.0 Features
 
+- `[Breadcrumb]` Added NG wrapper component for the IDS Breadcrumb API. `EPC` ([Issue #700](https://github.com/infor-design/enterprise-ng/issues/700))
 - `[Tree]` Added the ability to have separate icon button for expand/collapse and children count. ([#3847](https://github.com/infor-design/enterprise/issues/3847))
 - `[Datagrid]` Add summaryRow settings.  `AF`
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## v7.4.0
 
+### 7.4.0 Features
+
+- `[Breadcrumb]` Angular Wrapper Component for the IDS Breadcrumb API is now available. `EPC` ([Issue #700](https://github.com/infor-design/enterprise-ng/issues/700))
+
 ### 7.4.0 Fixes
 
 ## v7.3.0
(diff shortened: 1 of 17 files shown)


EdwardCoyle:
@tmcconechy The links on that test are just empty (not tied to any particular page or breadcrumb function).  Not sure why they'd be causing a page reload unless that's just what the app router will do.

tmcconechy:
Ok lets leave it for now and see what people say. In past had to preventDefault on the clicks. I noticed if you do pass in a href in the gauntet example it doesnt do that.

PR #891, opened 2020-08-10
889 - Remove Deprecated Log Message

### LINKED ISSUE #889: SohoModalDialogService: Regarding deprecation of parent argument
In recent versions of the components, the `parent` argument is marked as deprecated in the following API:
`public modal<T>(component: ComponentType<T>, parent?: ViewContainerRef, options?: SohoModalOptions)`

We would like to make sure that existing usages of `SohoModalDialogService.modal` continue to work, without having to change to a new call signature. Around 80 Homepages widgets are using this API in its current form, e.g. passing a ComponentType and a ViewContainerRef parent, as was required up until recently. If the parent argument is removed, as the deprecation states, it will probably break somewhere since the second argument would be treated as a SohoModalOptions object instead. 

Some alternatives that could be considered
* Don't remove the parent argument, keep it optional
    * The few (?) cases in which options are being passed, null could simply be provided as second argument. For others, ComponentType is sufficient 
* If it has to be removed, make sure to check if the second argument (if provided) is of type SohoModalOptions, not ViewContainerRef, before using as options.
    * The few cases in which options are being passed will "fail", since the options would not be used as there's no third argument anymore. Small code change required in the widget to make it work as before.

It's OK if TS compilation fails for the old syntax, just want to make sure that existing runtime JS will continue to work. 

**Additional context**
In `SohoContextualActionPanelService`, parent argument is not yet deprecated

tmcconechy:
Remove the message around deprecation for now. Even tho the property does nothing we cant commit to break it any time soon

Fixes #889

- re test modal 
- confirm API

### DIFF
--- a/projects/ids-enterprise-ng/src/lib/modal-dialog/soho-modal-dialog.service.ts
+++ b/projects/ids-enterprise-ng/src/lib/modal-dialog/soho-modal-dialog.service.ts
@@ -46,18 +46,14 @@ export class SohoModalDialogService {
    *
    * The dialog won't necessarily be open yet, see the dialog ref api for further methods.
    *
-   * @param component - the type of the component to instantiate; must not be null.
-   * @param parent - the parent container; obsolete and will be removed in a later release.
-   * @param options - the control options to use.
+   * @param component - The type of the component to instantiate; must not be null.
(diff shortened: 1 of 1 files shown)


tmcconechy:
cc @volante007

PR #908, opened 2020-09-10
Popupmenu: Expose is-selectable, is-multiselectable, and multi-selectable-section

### LINKED ISSUE #907: Popupmenu: Expose is-selectable, is-multiselectable, and multi-selectable-section
Popupmenu does not currently expose is-selectable, is-multiselectable, and multi-selectable-section.

Expose is-selectable, is-multiselectable, and multi-selectable-section as input properties.

crlang44:
Exposes is-selectable, is-multiselectable, and multi-selectable-section as input properties.

Closes #907 

1. Run the application. Go to http://localhost:4200/ids-enterprise-ng-demo/popupmenu.
2. Right click the first popupmenu.
3. Open the Multiselectable menu, see that each of the options can be checked or unchecked individually.


EdwardCoyle:
@crlang44 Thanks for this!  I was testing this and I noticed that the behavior of the submenus isn't quite right.  When you toggle an item in a Popupmenu's multiselectable list, the menu is supposed to stay open (unless the menu is specifically configured not to).  Examples of this can be seen [here](https://master-enterprise.demo.design.infor.com/components/popupmenu/example-selectable-multiple.html#) and [here](https://master-enterprise.demo.design.infor.com/components/popupmenu/example-selectable-multiple-with-headers.html#)

Did some digging and it seems that in the sample, when clicking on any of those four selectable items, the menu is immediately closed and the Popupmenu's internal method for [handling clicks](https://github.com/infor-design/enterprise/blob/master/src/components/popupmenu/popupmenu.js#L998), [selecting the anchor](https://github.com/infor-design/enterprise/blob/master/src/components/popupmenu/popupmenu.js#L1293) and eventually firing the [`selected` event](https://github.com/infor-design/enterprise/blob/55633e16f50c82cb8106fd25508f92ed1b415271/src/components/popupmenu/popupmenu.js#L1318) is bypassed.  To see this in action, you can set a breakpoint on that line in the complied `sohoxi.js` in the Angular demo, and set one in the [`close()` method](https://github.com/infor-design/enterprise/blob/master/src/components/popupmenu/popupmenu.js#L2321) too.

Not sure why `close()` is firing first but we should probably fix that before merging this.  The `handleItemClick()` method is primarily how the EP component manages its state internally so I suspect the current way toggling occurs might be throwing it off.

pwpatton:
We were having the same problem with dropdown multiselect and used `[closeOnSelect]="false"` to solve that.

crlang44:
image

The ids-ng demo for close on select just uses the default component so I think this is expected behavior.

tmcconechy (approved):

tmcconechy:
this ok @EdwardCoyle ?

EdwardCoyle (approved):

tmcconechy:
@crlang44 fyi i changed the default back to false for isSelectable https://github.com/infor-design/enterprise-ng/blob/7.6.2/projects/ids-enterprise-ng/src/lib/popupmenu/soho-popupmenu.component.ts#L81 as this caused an unintentional breaking change/effect of making menus selectable.

Just for your info

PR #911, opened 2020-09-16
E2e Cypress

nbcp:
This is a prototype to initiate writing of e2e tests using [Cypress ](https://www.cypress.io/) which provides us with a more human-friendly syntax and easier debugging. This contains tests from some of the commonly used patterns in landmark. This also includes the e2e in travis.yml to run these tests on our builds.

NA

1. `git checkout e2e-cypress`
2. `npm install`
3. `npm run test:e2e` <- runs in headless mode executing all tests, this is what's used in travis)
4. `npm run e2e:cy-debug` <- runs cypress in debug mode which requires a running serve, 
Debug mode shows the ff interface:
image
image


tmcconechy:
One test seemed to fail for me...

```
  1) Locale Tests
       should format data in English (British):
     AssertionError: Timed out retrying: Expected <div#dropdown-list.dropdown-list.placeable.is-closable> not to exist in the DOM, but it was continuously found.
      at Context.eval (http://localhost:4200/__cypress/tests?p=cypress/integration/locale/locale-format.spec.ts:115:18)
````

Overall i don't have a big problem including this as a POC for people to see if they wanted to use the test framework. The main issue i would have would be if we want to claim to support it / answer questions on it and do all the work on making tests for everything. We probably don't have resources for that level of support.

Also, maybe worth noting is i looked at this as well for a new project and decided to go with Jest + Puppeteer. For the following reasons:

- I didnt like the syntax of cypress as it was too different from what we were used to
- Since selecting Jest for functional tests adding jest-puppeteer seemed natural
- Cypress - pay model was making me wonder if one day we would loose it as open source or be forced into paying something for extra services
- Some of it is written in coffee script and using jQuery selectors (which we are getting away from)
- Seemed a bit slower and more opinionated

While we hadnt done too much e2e tests on the new work. Thats the direction im thinking there.

nbcp:
@tmcconechy 
> Also i couldn't get the debug/console to work? Do we have to run the sample app separate?

Yes the debug requires a running serve (or `npm start`), you'll notice the warning telling you that the localhost:4200 is not running

>The main question i would have would be who wants to be responsible for claiming we want to support it / answer questions on it and do all the work on making tests for everything. We probably don't have resources for that a high level of support.

I think this is wouldn't be a testing for everything, but just a testing for the most commonly used patterns the have known to be breaking a lot. We are planning two test engineers to be working on these. This being a part of the code review builds means it can safeguard against regressions that maybe caused by dependency updates that we may overlook via manual testing. @pwpatton may add more info on this

As for Jest, I think it is made for unit testing and not for e2e. I think this explains well about the most popular e2e options: https://medium.com/@sahil.goyal2/comparison-between-selenium-protractor-cypress-and-webdriverio-7786fc90ee09
I am really sold into the time travel and other debugging feature of cypress which the others lacks. I also think Cypress' test runner would be free for a very long time, their only paid option is when you want to use their dashboard service where your tests can upload videos, screenshots, and logs and view it in their fancy interface which I think we can live without.

tmcconechy:
Can you use screenshots for free or is that at cost? For e2e i'm using jest-puppeteer https://github.com/smooth-code/jest-puppeteer so this way you can stay using jest for both e2e and functional tests. 

I'm ok with letting this go in as an option but we probably won't be able to add much to it going forward. Lets see what others say...

pwpatton (approved):

tmcconechy (approved):

tmcconechy:
@nbcp - it seems this doesnt work on the CI server https://travis-ci.com/github/infor-design/enterprise-ng/builds/185206709 any idea why (need that to work or maybe we revert this back to what we had)

nbcp:
err.. I also encountered that issue. It is some sort of caching issue I haven't figured out yet how to fix in travis. The temp fix for it is to delete the cache in travis but it will come up again on the next run. We can disable the travis config for now in master until I can figure out how to deal with that cache issue.

tmcconechy:
i did try clearing the cache? Does the build need to run cypress install? Or perhaps we just make the build run the old tests for now?

nbcp:
@tmcconechy this one should fix it: #914 . Yeah `cypress install` is our only way to fix this now. It only adds 15secs to the build. I noticed from my other project using `npm ci`, this issue does not occur at all and no need to add anything to the travis config. Cypress recommends `npm ci` because `npm install` is known to be causing these cache issues. We can also get faster builds switching over to `npm ci` but we need to commit the package-lock file

PR #916, opened 2020-09-25
887/888 - Fix destroy and overwrite settings with context menu

### LINKED ISSUE #887: Contextmenu/Popupmenu not being destroyed/removed from DOM
**Describe the bug**
There doesn't seem to be a way to successfully remove a "**soho-context-menu**" and it's related "**soho-popupmenu**" from the view in Angular.

Using an **ngIf** directive to remove the "**soho-popupmenu**" component from the view leaves behind the generated "popupmenu" HTML elements in the DOM and, in many occasions, the popupmenu can still appear.

Even if we also set the **menu** @Input property to NULL of the "**soho-context-menu**", the wiring might still be left behind and the popupmenu can still appear. 

If we destroy the "**soho-context-menu**" component as well the mark-up is still left behind.

**To Reproduce**
1. Download: [context-menu.zip](https://github.com/infor-design/enterprise-ng/files/5042492/context-menu.zip)
2. Replace the files in the SoHo demo "src\app\context-menu" with the contents of the ZIP file.
3. Rebuild/Restart the SoHo demo.
4. Navigate to http://localhost:4200/ids-enterprise-ng-demo/context-menu

**Expected behavior**
Destroying the "**soho-popupmenu**" should remove all it's mark-up from the DOM. Additionally, setting the [menu] input of the "**soho-context-menu**" should remove the "popupmenu" from that element.

**Version**
- ids-enterprise-ng: I could reproduce this issue with slightly different results in **5.4.x**, **6.1.x** and **latest**.

**Screenshots**
Please, refer to this video to see the issue: [popupmenu-issue-01.zip](https://github.com/infor-design/enterprise-ng/files/5042528/popupmenu-issue-01.zip)

Notice: the button toggles a boolean which is used by the ngIf to remove the popupmenu component from the view and, to set the value of the menu property to NULL when false.

Moving the ngIf around might slightly change the behavior but, eventually you end up with the same results described above.

**Platform**
 - Device: Desktop
 - OS Version: Windows 10
 - Browser Name: Chrome
 - Browser Version: Latest

**Additional context**

This issue occurs even when you manually set the following properties:
    [attachToBody]="true"
    [removeOnDestroy]="true"

### LINKED ISSUE #888: Contextmenu/Popupmenu options (Inputs) overridden
**Describe the bug**
The "**soho-context-menu**" directive works with the "**soho-popupmenu**" component which is connected by the "**menu**" Input ("popupmenu" ID).

While debugging the "sohoxi.js" on Chrome, we noticed that this combination actually initializes twice:
1. One "**Popupmenu**" instance is created from the "**soho-context-menu**" directive.
2. A second "**Popupmenu**" instance is created from the "**soho-popupmenu**" component.

(The sequence changes if you change the elements order in the template but, the results are the same)

This seems to work and only one "popupmenu" is generated in the DOM...

However, the "**Popupmenu**" instance created by the "**soho-context-menu**" directive **DOES NOT** inherit/merge/use its assigned "**soho-popupmenu**" component options (or Inputs).

So, when the context-menu is opened, it uses default settings which means that, there is no way to configure the context-menu.

**To Reproduce**
1. Open src\app\context-menu\context-menu.demo.html
2. Copy/Page the code below:
```
<p soho-context-menu menu="popupmenu-issue-02">Right Click Here...</p>
<div id="parent-popupmenu-issue-02">
  <ul soho-popupmenu id="popupmenu-issue-02" [autoFocus]="false" [attachToBody]="false">
    <li soho-popupmenu-item *ngFor="let menuEntry of contextEntries" [isDisabled]="menuEntry.disabled">
      <a soho-popupmenu-label *ngIf="!menuEntry.url" menuId="{{menuEntry.id}}">{{menuEntry.displayString}}<span class="shortcut-text">{{menuEntry.shortcut}}</span></a>
      <a soho-popupmenu-label *ngIf="menuEntry.url" menuUrl="{{menuEntry.url}}">{{menuEntry.displayString}}<span class="shortcut-text">{{menuEntry.shortcut}}</span></a>
    </li>
  </ul>
</div>
```
3. Rebuild/Restart the SoHo demo.
4. Open http://localhost:4200/ids-enterprise-ng-demo/context-menu
5. Notice how the context-menu's "popupmenu" does not respect any of the options.

**Expected behavior**
The "**soho-context-menu**" should be able to use the options/inputs (settings) of its associated "**soho-popupmenu**" component.

**Version**
- ids-enterprise-ng: 5.4.x, 6.1.x, latest

**Screenshots**
Check this video of the debugging where the "Popupmenu" instances are created: 
[popupmenu-issue-02.zip](https://github.com/infor-design/enterprise-ng/files/5043264/popupmenu-issue-02.zip)
- First you can see that the "**soho-context-menu**" instance is created with the default values.
- Then you can see that the "**soho-popupmenu**" instance is created with the passed options.
- However, the context-menu does not respect the "popupmenu" options.

(Changing the other in the DOM will change the sequence but, it doesn't fix the problem).

**Platform**
 - Desktop
 - OS Version: Windows 10
 - Chrome
 - Latest

**Additional context**
In the last few releases there's been changes in what the default values of the settings should be and, in some cases, we need to overwrite those values however, it is not possible because the settings are not exposed in the directive and it is not using the ones from the component.

deep7102:
Fixed using *ngIf directive to destroy the component and overwrite api settings was not working properly with Context Menu.

Closes #887
Closes #888

Pull this branch and build/run the demo app
Navigate to: http://localhost:4200/ids-enterprise-ng-demo/context-menu-toggle

Issue: 887
- Use button `ShowContextMenu` to toggle
- Right click on the bottom text box to see the context menu
- True - should show custom context menu
- False - should show browser default context menu

Issue: 888
- Use button `autoFocus` to toggle (if disabled toggle `ShowContextMenu` to `true`)
- Right click on the bottom text box to open context menu
- True - should be focused first item
- False - should NOT be focused first item
- Open developer tools
- Use button `attachToBody` to toggle (if disabled toggle `ShowContextMenu` to `true`)
- Run in console `$('body > .popupmenu-wrapper').length;`
- True - should return `3`
- False - should return `2`

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### 7.7.0 Fixes
 
+- `[Context Menu]` Fixes a bug where using *ngIf directive to destroy the component was not working properly. ([#887](https://github.com/infor-design/enterprise-ng/issues/887))
+- `[Context Menu]` Fixes a bug where api settings were not overwriting the default settings. ([#888](https://github.com/infor-design/enterprise-ng/issues/888))
 - `[Datagrid]` Added support to disable column buttons. ([1590](https://github.com/infor-design/enterprise/issues/1590))
 - `[Popupmenu]` Expose is-selectable, is-multiselectable, and multi-selectable-section as input properties. ([#907](https://github.com/infor-design/enterprise-ng/issues/907)) `CL`
 
--- a/projects/ids-enterprise-ng/src/lib/context-menu/soho-context-menu.directive.ts
+++ b/projects/ids-enterprise-ng/src/lib/context-menu/soho-context-menu.directive.ts
@@ -2,21 +2,96 @@
 
 import {
   AfterViewInit,
+  Component,
   Directive,
   ElementRef,
   EventEmitter,
+  HostBinding,
   Input, NgZone,
   OnDestroy,
   Output
 } from '@angular/core';
 
+/**
+ * SUB COMPONENT: SOHO-CONTEXT-MENU-HEADING
+ */
+@Component({
+  selector: 'li[soho-context-menu-heading]', // tslint:disable-line
+  template: `<ng-content></ng-content>`
+})
+export class SohoContextMenuHeadingComponent {
+  @HostBinding('class.heading') get isHeading() { return true; }
+}
+
+/**
+ * SUB COMPONENT: SOHO-CONTEXT-MENU-SHORTCUT-TEXT
+ */
+@Component({
+  selector: 'span[soho-context-menu-shortcut-text]', // tslint:disable-line
+  template: `<ng-content></ng-content>`
+})
+export class SohoContextMenuShortCutTextComponent {
+  @HostBinding('class.shortcut-text') get isShortCutText() { return true; }
+}
+
+/**
+ * SUB COMPONENT: SOHO-CONTEXT-MENU-SEPARATOR
+ */
+@Component({
+  selector: 'li[soho-context-menu-separator]', // tslint:disable-line
+  template: `<ng-content></ng-content>`
+})
+export class SohoContextMenuSeparatorComponent {
+  @HostBinding('class.separator') get isSeparator() { return true; }
+  @HostBinding('class.single-selectable-section') @Input() singleSelectableSection = false;
+  @HostBinding('class.multi-selectable-section') @Input() multiSelectableSection = false;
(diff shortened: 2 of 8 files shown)


tmcconechy (approved):
Looks good to me. cc @edgarinfor is it possible for you to test this on your end?

PR #956, opened 2020-12-16
#955 stepprocess: feature to call selectStep when isSelected is set to true on one of the SohoStepListItemComponent

### LINKED ISSUE #955: Step process: when setting isSelected to true should select that panel into view
We navigate to another panel via code in the stepprocess component in enterprise-ng, we can only navigate to a panel via these three actions:
- clicking next
- clicking previous
- selecting the panel from the step list

Created a pull request when we set a SohoStepListItemComponent isSelected property to true, it will call the selectStep and should navigate to that panel and be shown into view.

None yet

**Additional context**
https://jira.lawson.com/browse/LMCLIENT-32916

JBismonte:

We cannot navigate to another panel via code, in this PR, we can just update the stepListItemComponent of that particular panel to true, and it will call selectStep to navigate and display that particular panel into view. 

- Closes #955 
- https://jira.lawson.com/browse/LMCLIENT-32916

- pull the code from a fork: https://github.com/JBismonte/enterprise-ng, build and run the demo app.
- go to http://localhost:4200/ids-enterprise-ng-demo/step-process-auto-step-change
- click any buttons, it should navigate to the correct panel.

See screen recording:

https://user-images.githubusercontent.com/29673378/102778777-25b6e800-43ce-11eb-8334-5a5fadb60231.mp4

note: 
- when you set the isSelected to true in a panel that is a folder, it will toggle the folder to either open or close.

### DIFF
--- a/projects/ids-enterprise-ng/src/lib/stepprocess/soho-stepprocess.component.ts
+++ b/projects/ids-enterprise-ng/src/lib/stepprocess/soho-stepprocess.component.ts
@@ -77,7 +77,7 @@ export class SohoStepListItemComponent {
 
   @HostBinding('class.is-selected') @Input() set isSelected(isSelected: boolean) {
     if (isSelected) {
-      let stepLink = $(this.element.nativeElement).children('a.js-step-link');
+      const stepLink = $(this.element.nativeElement).children('a.js-step-link');
       this.sohoStepProcessComponent.stepprocess?.selectStep(stepLink);
     }
   }
--- a/projects/ids-enterprise-ng/src/lib/stepprocess/soho-stepprocess.component.ts
+++ b/projects/ids-enterprise-ng/src/lib/stepprocess/soho-stepprocess.component.ts
@@ -82,6 +82,8 @@ export class SohoStepListItemComponent {
     }
   }
 
+  @Input() stepId?: string;
+
   constructor(@Host() private sohoStepProcessComponent: SohoStepProcessComponent, private element: ElementRef) { }
 }
 
--- a/projects/ids-enterprise-ng/src/lib/stepprocess/soho-stepprocess.component.ts
+++ b/projects/ids-enterprise-ng/src/lib/stepprocess/soho-stepprocess.component.ts
@@ -75,20 +75,15 @@ export class SohoStepListItemComponent {
     return true;
   }
 
+  @HostBinding('class.is-selected') _isSelected = false;
+
   @Input() set isSelected(isSelected: boolean) {
+    this._isSelected = isSelected;
     if (isSelected) {
       const stepLink = $(this.element.nativeElement).children('a.js-step-link');
       this.sohoStepProcessComponent.stepprocess?.selectStep(stepLink);
     }
-    this._isSelected = isSelected;
   }
-
-  get isSelected() {
-    return this._isSelected;
-  }
-
-  @HostBinding('class.is-selected') _isSelected: boolean = false;
-
 
   constructor(@Host() private sohoStepProcessComponent: SohoStepProcessComponent, private element: ElementRef) { }
 }
--- a/projects/ids-enterprise-ng/src/lib/stepprocess/soho-stepprocess.component.ts
+++ b/projects/ids-enterprise-ng/src/lib/stepprocess/soho-stepprocess.component.ts
@@ -75,14 +75,20 @@ export class SohoStepListItemComponent {
     return true;
   }
 
-  @HostBinding('class.is-selected') @Input() set isSelected(isSelected: boolean) {
+  @Input() set isSelected(isSelected: boolean) {
     if (isSelected) {
       const stepLink = $(this.element.nativeElement).children('a.js-step-link');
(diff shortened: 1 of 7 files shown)


tmcconechy (dismissed):

nbcp (changes requested):
@JBismonte please also include your changes for the demo page here. That way it can be used for testing this fix and future use cases.

src/app/stepprocess/stepprocess-auto-step-change.demo.ts
```diff
@@ -0,0 +1,105 @@
...
+    setTimeout(() => {
+      if (this.stepListItems) {
+        const stepListItemsLength = this.stepListItems.toArray().length;
+        const stepNo = Math.floor(Math.random() * stepListItemsLength);
+        this.stepListItems.toArray()[stepNo].isSelected = true;
+        console.log(stepNo);
+        console.log('step changed');
+      }
+    }, 11000);
+
```
nbcp, line 99: lets make it just select one defined random sub step, this way we can add an automated test for it in the future.

nbcp: Actually there's no need for this and the ViewChildren since isSelected is an Input in SohoStepListItemComponent, we can just add `[isSelected]="true"`  to one of the `soho-step-list-item` in the template

nbcp (changes requested):

src/app/stepprocess/stepprocess-auto-step-change.demo.ts
nbcp: I just realize http://localhost:4200/ids-enterprise-ng-demo/step-process already does this so maybe we can just use that

JBismonte: Hi @nbcp, tried to refactor the code of the demo page, and it will rely on the clicking of the buttons rather than awaiting the setTimeout functions in the ngAfterViewInit.

projects/ids-enterprise-ng/src/lib/stepprocess/soho-stepprocess.component.ts
```diff
@@ -82,6 +82,8 @@ export class SohoStepListItemComponent {
     }
   }
 
+  @Input() stepId?: string;
```
nbcp, line 85: why was this added? please remove it if it is only needed for the demo component

src/app/stepprocess/stepprocess-auto-step-change.demo.html
```diff
@@ -8,21 +8,21 @@
 
     <ng-template ngFor let-step [ngForOf]="steps">
 
-      <li soho-step-list-item *ngIf="step.content">
+      <li soho-step-list-item stepId="{{step.id}}" *ngIf="step.content">
```
nbcp, line 11: why go adding a new input when you could just do something like:
`[isSelected]={{step.id == selectedStepId}}`

src/app/stepprocess/stepprocess-auto-step-change.demo.ts
```diff
@@ -36,70 +38,23 @@ export class StepProcessAutoStepChangeComponent implements OnInit, AfterViewInit
...
-        console.log('step changed');
-      }
-    }, 11000);
 
+  selectPanel(id: string) {
+    console.log(id);
+    const selectedItem = this.stepListItems.find((item) => item.stepId === id);
+    if (selectedItem) {
+      selectedItem.isSelected = true;
+    }
```
nbcp, line 52: and here your could just:
```suggestion
    this.selectedStepId = id;
```

nbcp (changes requested):

projects/ids-enterprise-ng/src/lib/stepprocess/soho-stepprocess.component.ts
```diff
@@ -73,7 +74,17 @@ export class SohoStepListItemComponent {
...
     return true;
   }
-  @HostBinding('class.is-selected') @Input() isSelected = false;
+
+  @HostBinding('class.is-selected') @Input() set isSelected(isSelected: boolean) {
+    if (isSelected) {
+      const stepLink = $(this.element.nativeElement).children('a.js-step-link');
+      this.sohoStepProcessComponent.stepprocess?.selectStep(stepLink);
+    }
+  }
```
nbcp, line 83: Since this becomes a setter, we'll also need a to set the value in a private property and expose it via a getter. Hostbinding for class should be moved there as well

bthharper: Agree with the above comment.

```typescript 

  @HostBinding('class.is-selected') _isSelected: boolean = false;

  @Input() set isSelected(isSelected: boolean) {
    this._isSelected = isSelected;
    if (isSelected) {
      const stepLink = $(this.element.nativeElement).children('a.js-step-link');
      this.sohoStepProcessComponent.stepprocess?.selectStep(stepLink);
    }
  }
```

nbcp (approved):

PR #982, opened 2021-02-03
SohoRatingComponent expose val method

### LINKED ISSUE #981: Rating: expose the val() method
With the change to rating to allow stars to be unselected (https://github.com/infor-design/enterprise/issues/4571), we need a way to get the current value of the control. Exposing the val() method will allow this, and allow programmatic setting of the control as well.

Expose the val() method from enterprise controls.

Sovia:
Exposing the val method in Rating component

Closes #981

Pull in change, navigate to the rating page in the demo app.
1. Enable the control
2. Click the stars and see that the text under the control gets updated to the correct current value

### DIFF
--- a/projects/ids-enterprise-ng/src/lib/rating/soho-rating.component.ts
+++ b/projects/ids-enterprise-ng/src/lib/rating/soho-rating.component.ts
@@ -52,4 +52,19 @@ export class SohoRatingComponent implements AfterViewInit, OnDestroy {
       this.rating.enable();
     }
   }
+
+  /** Gets the current value of the control. If the control doesn't exist, will return -1 */
+  public get currentValue() {
+    if (this.rating) {
+      return this.rating.val();
+    }
+    return -1;
+  }
+
+  /** Sets the value for the control */
+  public set currentValue(newValue: number) {
+    if (this.rating) {
+      this.rating.val(newValue);
+    }
+  }
 }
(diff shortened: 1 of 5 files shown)


pwpatton (approved):

projects/ids-enterprise-ng/src/lib/rating/soho-rating.component.ts
```diff
@@ -52,4 +52,19 @@ export class SohoRatingComponent implements AfterViewInit, OnDestroy {
       this.rating.enable();
     }
   }
+
+  /** Gets the current value of the control. If the control doesn't exist, will return -1 */
+  public get currentValue() {
```
tmcconechy, line 57: Do we want to call it val() in both places? I actually prefer currentValue so ok with leaving this for the public API but most of the time we stick to the name...

projects/ids-enterprise-typings/lib/rating/soho-rating.d.ts
```diff
@@ -29,6 +29,9 @@ interface SohoRating {
 
   /** Destroys the control on completion. */
   destroy(): void;
+
+  /** Returns the current value if called with no value, or sets the current value to the passed in value*/
+  val(newValue?: number): number;
```
tmcconechy, line 34: I think this doesnt match the val vs currentValue

Issue #993, opened 2021-02-17
Contextual Action Panel: close() API to support same behavior as jQuery counterpart


volante007:
**Is your feature request related to a problem? Please describe.**
This is related to the below issue, which was fixed for the enterprise controls, regarding needing to click twice to close the CAP when a child comp. is opened:
https://github.com/infor-design/enterprise/issues/4112

The solution was adding an optional parameter to the CAP's `close()` API, e.g.
https://github.com/infor-design/enterprise/pull/4221/files#diff-24c310eed7375baa9d4ae68f49bb7e170ddddf774da4bf3778edea480647b0f9R514

**Describe the solution you'd like**
Update the `close()` function in `SohoContextualActionPanelRef` to support the same `doForce` flag. Currently it looks like:
```
 /**
   * Closes the panel panel, if open.  The panel is not closed
   * fully until the 'afterClosed' event is fired.
   *
   * @param panelResult - optional result - passed back to the caller.
   */
  close(panelResult?: any): SohoContextualActionPanelRef<T> {
    this.panelResult = panelResult;
    if (this.contextualactionpanel) {
      this.ngZone.runOutsideAngular(() => {
        this.contextualactionpanel?.close();
      });
    }
    return this;
  }
```
Would need to be something like:
```
 /**
   * Closes the panel panel, if open.  The panel is not closed
   * fully until the 'afterClosed' event is fired.
   *
   * @param panelResult - optional result - passed back to the caller.
   */
  close(panelResult?: any, doForce?: boolean): SohoContextualActionPanelRef<T> {
    this.panelResult = panelResult;
    if (this.contextualactionpanel) {
      this.ngZone.runOutsideAngular(() => {
        this.contextualactionpanel?.close(doForce);
      });
    }
    return this;
  }
``` 

**Additional context**
It seems though that the original fix should "automatically" support the scenario of opened child comps. when clicking close, but it doesn't work when using the Angular version, so I guess the doForce flag needs to be passed.

PR #1007, opened 2021-03-15
993 - update close() api for CAP to force close

ericangeles:

Update the close() API to behave the same way as the jQuery version of CAP. Added an optional param `doForce` to force the CAP to close.

Closes https://github.com/infor-design/enterprise-ng/issues/993

- Go to http://localhost:4200/ids-enterprise-ng-demo/contextual-action-panel
- Click `Panel with Searchfield in flex toolbar` button to trigger the CAP
- Click the dropdown `Menu` to open it
- Then close the modal
- It should close the modal immediately


tmcconechy (approved):

volante007:
Hi, a suggestion for the change. Shouldn't the new parameter be appended rather than prepended, to avoid breaking existing usages? Perhaps someone is passing a boolean as a dialog result (which now becomes doForce value instead). Compilation will not hint that something is wrong.

tmcconechy:
cc @ericangeles what do you think?

ericangeles:
@tmcconechy @volante007 yeah that does make sense. Let me change the approach.

PR #1015, opened 2021-03-26
Closes #1014

### LINKED ISSUE #1014: Need to update modal API and to be able to use force parameter to cl is an opened tooltip
**Describe the bug**
Enterprise's modal has this ff condition in close method:
```
  close(destroy, noRefresh, force = false) {
    if (!force && (!this.visible || this.openSubComponents.length)) {
      return true;
    }
```
Calling modal.close if there is an open tooltip won't close the modal because this.openSubComponents.length becomes true.

**To Reproduce**
This is a scenario where I can only replicate in landmark where after clicking a form save button, the button get's focused and then later shows a tooltip after showing more than 2 modals. This said tooltip however is being detected in enterprise via `this.openSubComponents.length` which prevents closing the modal. I notice there is a force in the close API we can use to override that condition and be able to proceed with close.

**Expected behavior**
Need to be able to do a force close to a modal when close is explicitly called.

**Version**
- ids-enterprise-ng: 9.3.1

nbcp:
This updates the SohoModalStatic interface to accept enterprise's close parameters and defaults force parameter into true. This ensures the modal closes when modal.close() is called even when there is a open tooltip.

closes #1014

Visual inspect, this change is trivial

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -6,6 +6,12 @@
 
 - `[Placeholder]` Remove me.... ([#1001](https://github.com/infor-design/enterprise-ng/issues/1001)) `TJM`
 
+## v9.3.2
+
+### 9.3.2 Fixes
+
+- `[Modal]` Update modal close method API and use force as default to proceed modal closing even when tooltip is open. `NBCP` ([#1014](https://github.com/infor-design/enterprise-ng/issues/1014))
+
 ## v9.3.1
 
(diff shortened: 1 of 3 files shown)


tmcconechy (dismissed):
Change log maybe?

nbcp:
> Change log maybe?

Updated changelog, can we make a patch v9.3.2 please? thanks

ericangeles (approved):

PR #1022, opened 2021-03-31
getValue returns string or Date object.

### LINKED ISSUE #1016: DatePicker: getValue returns string instead of Date object
**Describe the bug**
Calling getValue from SohoDatePickerComponent returns the selected date as string. The method is defined as `public getValue(): Date` and should return a value with type Date.

**To Reproduce**
1. Change the source code in enterprise-ng/src/app/datepicker/datepicker.demo.ts and add a new console output in onChange method: `console.log('DatePickerDemoComponent.onChange(date of datepicker): ' + this.datepicker?.getValue());`.
2. Start the application with `npm start`.
3. Open the accordion with letter D on the left side and choose Date Picker.
4. Select a different value in Standard Date field to trigger a change event.
5. Open developer console window in your browser.
6. See the log message like `DatePickerDemoComponent.onChange(date of datepicker): 12/13/2016`.

**Expected behavior**
In general I expect a Date object after calling getValue instead of string. The steps "To Reproduce" should end with the following log message: `DatePickerDemoComponent.onChange(date of datepicker): [object Object]`. And the type of object should be Date to call according methods on it. 

**Version**
- ids-enterprise-ng: 9.1.5

**Screenshots**
no screenshots

**Platform**
 - Device (if applicable): Mac mini
 - OS Version: macOS Big Sur 11.2.3
 - Browser Name: safari
 - Browser Version: Version 14.0.3 (16610.4.3.1.7)

**Additional context**
no additional context

swuendri:
The DatePicker component now returns the selected date as string or Date object.

Fixes #1016

- getValue method was changed according the discussion in issue
- 3 additional automated tests check the new behavior
- for test 'Check getting value with time' one remark: setting a date with 6 o'clock results in 12 o'clock.

### DIFF
--- a/projects/ids-enterprise-ng/src/lib/datepicker/soho-datepicker.component.ts
+++ b/projects/ids-enterprise-ng/src/lib/datepicker/soho-datepicker.component.ts
@@ -386,7 +386,17 @@ export class SohoDatePickerComponent extends BaseControlValueAccessor<any> imple
     this.datepicker?.setValue(value, true);
   }
 
-  public getValue(): Date {
+  public getValue(asDate: boolean = false): string | Date {
+    if (asDate) {
+      const calendar = Soho.Locale.calendar();
+      const dateFormat = this._options.dateFormat || calendar.dateFormat.short;
+      const timeFormat = this._options.timeFormat || calendar.timeFormat;
+      let format = dateFormat;
+      if (this._options.showTime) {
+        format += 'T' + timeFormat;
+      }
+      return Soho.Locale.parseDate(this.internalValue, format);
+    }
     return this.internalValue;
   }
 
--- a/projects/ids-enterprise-ng/src/lib/datepicker/soho-datepicker.spec.ts
+++ b/projects/ids-enterprise-ng/src/lib/datepicker/soho-datepicker.spec.ts
@@ -381,4 +381,78 @@ describe('Soho Datepicker Unit Tests', () => {
       });
     });
   }));
+
+  it('Check getting value for en-US', waitForAsync(() => {
+    spyOn(comp, 'onChange');
+
+    const date = new Date('1978-11-11T12:00:00Z');
+    const dateWithoutTime = getDateWithoutTime(date);
+
+    expect(Soho.Locale.currentLocale.name).toEqual('en-US');
+    comp.datepicker?.setValue(date);
+
+    fixture.detectChanges();
+
+    fixture.whenStable().then(() => {
+      fixture.detectChanges();
+      expect(Soho.Locale.currentLocale.name).toEqual('en-US');
+      expect(comp.onChange).toHaveBeenCalled();
+      expect(['11/11/1978', '1978-11-11']).toContain(comp.datepicker?.getValue() as any);
+      expect(dateWithoutTime).toEqual(comp.datepicker?.getValue(true) as any);
+    });
+  }));
+
+  it('Check getting value for different date format', waitForAsync(() => {
+    comp.dateFormat = 'dd:MM:yyyy';
+    fixture.detectChanges();
+
+    spyOn(comp, 'onChange');
+
+    const date = new Date('1978-11-11T12:00:00Z');
+    const dateWithoutTime = getDateWithoutTime(date);
+
+    expect(Soho.Locale.currentLocale.name).toEqual('en-US');
+    comp.datepicker?.setValue(date);
+
+    fixture.detectChanges();
+
+    fixture.whenStable().then(() => {
+      fixture.detectChanges();
+      expect(Soho.Locale.currentLocale.name).toEqual('en-US');
+      expect(comp.onChange).toHaveBeenCalled();
+      expect('11:11:1978').toBe(comp.datepicker?.getValue() as any);
+      expect(dateWithoutTime).toEqual(comp.datepicker?.getValue(true) as any);
+    });
+  }));
+
+  it('Check getting value with time', waitForAsync(() => {
+    comp.showTime = true;
+    fixture.detectChanges();
+
+    spyOn(comp, 'onChange');
+
+    const date = new Date('1978-11-11T06:00:00Z');
+    const dateWithoutTime = getDateWithoutTime(date);
+
+    expect(Soho.Locale.currentLocale.name).toEqual('en-US');
+    comp.datepicker?.setValue(date);
+
+    fixture.detectChanges();
+
+    fixture.whenStable().then(() => {
+      fixture.detectChanges();
+      expect(Soho.Locale.currentLocale.name).toEqual('en-US');
+      expect(comp.onChange).toHaveBeenCalled();
+      expect('11/11/1978 12:00 AM').toBe(comp.datepicker?.getValue() as any);
+      expect(dateWithoutTime).toEqual(comp.datepicker?.getValue(true) as any);
+    });
+  }));
+
+  function getDateWithoutTime(date: Date): Date {
+    const dateWithoutTime = date;
+    dateWithoutTime.setHours(0);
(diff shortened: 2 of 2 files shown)


tmcconechy (approved):
Looks good - thanks for your contribution ! FYI I fixed the checks as only 14x are now needed

tmcconechy (approved):

tmcconechy:
@swuendri do you think it would work for your case if you revised the PR based on a conversation here https://teams.microsoft.com/l/message/19:83173693057f4991b7a84f030975c06d@thread.skype/1617638545677?tenantId=457d5685-0467-4d05-b23b-8f817adda47c&groupId=4f50ef7d-e88d-4ccb-98ca-65f26e57fe35&parentMessageId=1617638545677&teamName=IDS%20Enterprise%20Development%20(SoHo%20Xi)&channelName=Angular%20Components&createdTime=1617638545677


perhaps
```
  public getValue(asDate: boolean = false): string | Date {
    const pickerAPI = (this.datepicker as any);
    if (asDate && !(pickerAPI.currentDate || pickerAPI.currentDateIslamic )) {
      const calendar = Soho.Locale.calendar();
      const dateFormat = this._options.dateFormat || calendar.dateFormat.short;
      const timeFormat = this._options.timeFormat || calendar.timeFormat;
      let format = dateFormat;
      if (this._options.showTime) {
        format += 'T' + timeFormat;
      }
      return Soho.Locale.parseDate(this.internalValue, format);
    }
    if (asDate && (pickerAPI.currentDate || pickerAPI.currentDateIslamic )) {
       const pickerAPI = (this.datepicker as any);
       return pickerAPI.isIslamic ? pickerAPI.currentDateIslamic : pickerAPI.currentDate;
    }
    return this.internalValue;
  }
```

This way it would work for partial dates (month/year) and return an array for islamic dates

swuendri:
@tmcconechy I'm not able to read teams messages but I think to understand the recommended change. I will add autmated test(s) for partial dates and do the change because we also use this in our project. With islamic dates I'm currently not familiar so I will try my best.

tmcconechy:
Actually the problem was if the Date is without days fx May 2020 - then using the method with using Locale.parseDate didnt work. So this is changing it to use an internal date object.

The problem is this is only populated on selection but for that case this is ok. So using the string parse as a fallback

swuendri:
Sorry @tmcconechy but currently I cannot produce an error with month/year picker. In my automated test I get the first day of a month as result which is convert able into a Date object. Are there examples in Teams thread which doesn't work? And can you provide these examples here?

tmcconechy:
thats ok @swuendri - so it should be an example where only part of the date is shown. So any of theses

https://master-enterprise.demo.design.infor.com/components/datepicker/example-month-year-format.html

Thinking about it it might just set the day to 1? If thats the case maybe thats the best we can do?

swuendri:
It tooks a while because I had to thing about behavior before and the expected changes after introducing asDate option. I have created a table with all possible options and the influence to getValue(true|false). Please review my assumptions and let us find the right expectation. Afterwards I will create all pull request with necessary changes.

The table shows the option first. The next two columns describe the behavior of getValue with asDate false and true. The following two columns are informative and describe the type and content of the option.

I used the '-' character to say, that there is no influence to the result of getValue by this option. The 'no change' comment means that the behavior is unchanged like before introducing the option asDate.

I think all other comments needs no description. But please don't hesitate to ask question.

Before creating this table I checked the behavior with enterprise-ng datepicker demo. In my opinion options showMonthYearPicker and hideDays doesn't work like expected. But I'm not sure if my understanding is correct. But I would create Issues if you have same opinion.

|Name of option                    |Expected getValue(false) result|Expected getValue(true) result|Type of option             |Description                                                                                                        |
|----------------------------------|-------------------------------|------------------------------|---------------------------|-------------------------------------------------------------------------------------------------------------------|
|showTime                          |no change                      |Date with Time                |boolean                    |Include the time in the date picker; defaults to false (in the soho control).                                      |
|useCurrentTime                    |no change                      |Date with Time                |boolean                    |If true current time will be used for the time portion otherwise 12:00 midnight is used.                           |
|timeFormat                        |no change                      |Date with Time                |string                     |Time format, e.g. HH:mm:ss.                                                                                        |
|minuteInterval                    |-                              |-                             |number                     |An integer from 1 to 60; multiples of this value are displayed as options in the minutes dropdown.                 |
|secondInterval                    |-                              |-                             |number                     |An integer from 1 to 60; multiples of this value are displayed as options in the seconds dropdown.                 |
|mode: standard                    |-                              |-                             |SohoDatePickerMode         |Indicates the entry mode.                                                                                          |
|mode: range                       |no change                      |undefined                     |"                          |"                                                                                                                  |
|roundToInterval                   |-                              |-                             |boolean                    |If a non-matching minutes value is entered, will round the minutes value to the nearest interval on the blur event.|
|dateFormat: with year, month, date|no change                      |Date with default Time        |string                     |The pattern used to format the date; or a locale to use.                                                           |
|dateFormat: with year, month      |no change                      |undefined                     |"                          |"                                                                                                                  |
|dateFormat: with month, date      |no change                      |undefined                     |"                          |"                                                                                                                  |
|placeholder                       |-                              |-                             |string                     |Display a placeholder for empty values?                                                                            |
|disable                           |-                              |-                             |SohoDatePickerDisable      |A date or range of dates that are enabled/disabled.                                                                |
|showLegend                        |-                              |-                             |boolean                    |If true a legend is show to associate dates.                                                                       |
|showMonthYearPicker               |no change                      |Date with default Time        |boolean                    |If true the month and year will render as dropdowns.                                                               |
|hideDays                          |no change                      |Date with default Time        |boolean                    |If true the days portion of the calendar will be hidden.                                                           |
|yearsAhead                        |-                              |-                             |number                     |The number of years ahead to show in the month/year picker should total 9 with yearsBack.                          |
|yearsBack                         |-                              |-                             |number                     |The number of years back to show in the month/year picker should total 9 with yearsAhead.                          |
|legend                            |-                              |-                             |Array\<SohoDatePickerLegend>|The number of months in each direction to show in the dropdown for months (when initially opening)                 |
|range                             |-                              |-                             |SohoDatePickerRange        |Use range of two dates options.                                                                                    |
|calendarName: gregorian           |no change                      |Date with default Time        |SohoDatePickerCalendarName |Calendar name.                                                                                                     |
|calendarName: islamic-umalqura    |no change                      |Date with default Time        |"                          |"                                                                                                                  |
|locale                            |-                              |-                             |string                     |The name of the locale to use for this instance. If not set, the current locale will be used.                      |
|useUTC                            |-                              |-                             |boolean                    |If true the dates will use UTC format. This is only partially implemented https://jira.infor.com/browse/SOHO-3437  |
|firstDayOfWeek                    |-                              |-                             |SohoDatePickerDayOfWeek    |Set first day of the week. '1' would be Monday.                                                                    |
|autoSize                          |-                              |-                             |boolean                    |If true the field will be sized to the width of the date.                                                          |
|showToday                         |-                              |-                             |boolean                    |Show the today button on the header.                                                                               |
|attributes                        |-                              |-                             |Array\<Object> &#124; Object     |Add extra attributes like id's to the component                                                                    |

swuendri:
@tmcconechy, had you time to think about my last post? Do we want to continue this topic?

tmcconechy:
@swuendri i'm not following that post (and sorry i missed it). But i think the only thing presently missing that im aware of is that if you have an example like https://master-enterprise.demo.design.infor.com/components/datepicker/example-month-year-format.html then maybe it needs to default the day to zero in place that there isnt a day. It may already be doing that?

But we also could just leave this for now as im not sure there is any open issue. So lets leave this for now...

swuendri:
Okay @tmcconechy but if there is an issue you can remember me.

PR #1025, opened 2021-04-05
1024 - Fix switching datepicker mode wasn't setting the range object in options correctly

### LINKED ISSUE #1024: Field Filter: Datepicker remains in range mode after resetting to a field filter of equals
**Describe the bug**
When using a field filter, if you set a date picker to In Range and then back to Equals, when you open the date picker it remains in  In Range mode.

I believe this is because the mode is getting reset in for the options object, but the range: {useRange: true} is staying there. When setting back to standard mode, we need to clear the range object in the options.

**To Reproduce**
1. In the angular demo app, go to F -> Field Filter
2. Scroll down to Date With Range, and set the filter field to In Range
3. Set the filter field to Equals
4. Open the date picker and see that the picker remains in range mode.

**Expected behavior**
The date picker should have switched back to a single date select mode.

**Version**
- ids-enterprise-ng: v4.38.0-dev.20210221

**Screenshots**
DateRemainingInRangeMode

**Platform**
 - All

**Additional context**

Sovia:
Sets the datepicker options correctly when switching back from range mode.

Closes #1024 

1. Pull in my code change
2. In the angular demo app, go to F -> Field Filter
3. Scroll down to Date With Range, and set the filter field to In Range.
4. Open the picker and validate it is now in range mode
4. Set the filter field to Equals
5. Open the date picker and see that the picker is now in single select mode.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - `[Charts]` Added double click event to all chart types. `DV` ([#3229](https://github.com/infor-design/enterprise/pull/3229))
 - `[Popupmenu]` Add example demonstrating shared Popupmenus working after one is destroyed. ([#987](https://github.com/infor-design/enterprise-ng/issues/987)) `EPC`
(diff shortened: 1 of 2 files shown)


Sovia:
Landmark will need this patched back to 9.3, right @vonnyw ?

tmcconechy (approved):

Issue #1107, opened 2021-08-17
Searchfield: Add getCategoryData method


whernebrink:
Expose `getCategoryData(onlyReturnSelected)` on the searchfield component.

Enterprise has [method](https://github.com/infor-design/enterprise/blob/6368aa74b9dd876091d7bb09e4c36b8f9a4e9258/src/components/searchfield/searchfield.js#L1587) we'd like to call:

We have tried the `getSelectedCategories()` but that returns jQuery list so a bit cumbersome to use.

tmcconechy:
Feel free to do a PR if we arent able to get to this soon

PR #1108, opened 2021-08-18
1107 - Add missing searchfield function and improve typings

whernebrink:
Adding `getCategoryData()` to the api facilitates getting the actual categories. There is already a `getSelectedCategories()` but that one returns a list of jQuery items, which then makes it cumbersome to work with.

Also, to facilitate working with the categories, a `SohoSearchFieldCategory` interface (which corresponds to what you'd get back from the `getCategoryData()`)  was created, as well as a `SohoSearchFieldCategoryType` which is a what you can pass to the `searchfieldOptions.categories`, which is an array of strings and/or `SohoSearchFieldCategory` items. Previously this was typed to `categories: Object[];`.

Fixes https://github.com/infor-design/enterprise-ng/issues/1107

- Pull this branch and run the build for typings and lib
- Run the demo app
- Navigate to: http://localhost:4200/ids-enterprise-ng-demo/searchfield-category
- When selecting any categories, these should be seen in the ui.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### 10.4.0 Fixes
 
 - `[Lookup]` Added autocomplete in lookup. ([#1087](https://github.com/infor-design/enterprise/issues/1087))
+- `[SearchField]` Added method for getting category data and improved typings for searchfield cateegories. ([#1107](https://github.com/infor-design/enterprise-ng/issues/1107))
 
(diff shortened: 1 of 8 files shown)


tmcconechy (dismissed):
Thanks @whernebrink  !

whernebrink:
@tmcconechy Accidentally commented the categories array so I uncommented that and cleaned up the example a bit.

Issue #1215, opened 2022-01-19
CAP: cssClass not added to modal element


whernebrink:
**Describe the bug**
Not able to apply cssClass to CAP modal.

**To Reproduce**
1. Go to this [this stackblitz](https://stackblitz.com/edit/ids-quick-start-1301-nwg4ig?file=src%2Fapp%2Fapp.component.html)
2. Click on the "Show Cap Dialog" button
3. Inspect the DOM
4. You will see that there is no added class to the element:
![image](https://user-images.githubusercontent.com/73601152/150152614-a215a795-3844-4d8b-b51e-5e2558df7f10.png)

5. Close the modal and click the "Open Standard Soho Modal"
6. Inspect the DOM and you will see the added class name:
![image](https://user-images.githubusercontent.com/73601152/150153244-020789aa-cfc5-4c72-9d80-c39309ff4214.png)

**Expected behavior**
Using the `.cssClass('my-custom-class')` or `.modalSettings({ cssClass: 'my-custom-class'})` should be applied to the element with the `.modal` class.

**Version**
- ids-enterprise-ng: 13.0.0

**Platform**
 - OS Version: Mac OSX
 - Browser Name: n/a
 - Browser Version n/a

whernebrink:
Some additional details to this.

This works (first opening the modal, then apply the cssClass do the dialogRef), which is kinda weird and differs from how we usually use the soho modal one:
<img width="566" alt="image" src="https://user-images.githubusercontent.com/73601152/150412246-f062c440-ed03-49a5-b1c3-1ce43b980bf6.png">

Seems to be because of the following code in `soho-contextual-action-panel.ref.ts` wich checks to see that the dialog is open:
<img width="1352" alt="image" src="https://user-images.githubusercontent.com/73601152/150412535-a42dde49-3cdc-40ae-971b-5ba225fd9308.png">

jbrcna:
this issue is now resolved.
verified in https://main-enterprise.demo.design.infor.com/components/contextualactionpanel/test-css-settings.html
<img width="1244" alt="image" src="https://user-images.githubusercontent.com/56722184/156337141-17855e4b-b1c0-454a-bacd-05b9fedda372.png">

PR #1242, opened 2022-02-28
1215 - CAP cssClass setting

tjamesallen15:

This PR will added setting for cssClass option in CAP.

Closes https://github.com/infor-design/enterprise-ng/issues/1215

- Go to http://localhost:4200/ids-enterprise-ng-demo/contextual-action-panel
- Click the button `Open Panel CSS`
- Open `Developer Console`
- Inspect and Check to see `my-custom-panel-test` inside the CAP

**Included in this Pull Request**:
- [ ] An e2e or functional test for the bug or feature.
- [x] A note to the change log.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -4,9 +4,9 @@
 
 ### 13.2.0 Fixes
 
-- `[Datagrid]` Added missing type for `doNotEmptyCellWhenEditing`. ([#5849](https://github.com/infor-design/enterprise/issues/5849))
+- `[ContextualActionPanel]` Added setting for cssClass in CAP. ([#1215](https://github.com/infor-design/enterprise-ng/issues/1215))
 - `[Context Menu]` Context menu should appear even when dataset is updated in listview. ([#1119](https://github.com/infor-design/enterprise-ng/issues/1119))
-
+- `[Datagrid]` Added missing type for `doNotEmptyCellWhenEditing`. ([#5849](https://github.com/infor-design/enterprise/issues/5849))
 - `[Datagrid]` Fixed removeRow typing ([#1238](https://github.com/infor-design/enterprise-ng/issues/1238))
 
 ## v13.1.1
--- a/projects/ids-enterprise-ng/src/lib/contextual-action-panel/soho-contextual-action-panel.ref.ts
+++ b/projects/ids-enterprise-ng/src/lib/contextual-action-panel/soho-contextual-action-panel.ref.ts
@@ -131,6 +131,11 @@ export class SohoContextualActionPanelRef<T> {
         const cap = jQuery('div.contextual-action-panel.modal');
         cap.addClass(`${cssClass}`);
       });
+    } else {
+      this._options = $.extend(true, this._options, { cssClass: cssClass });
+      if (this.contextualactionpanel) {
+        this.contextualactionpanel.settings = this._options;
+      }
     }
 
     return this;
--- a/projects/ids-enterprise-typings/lib/contextual-action-panel/soho-contextual-action-panel.d.ts
+++ b/projects/ids-enterprise-typings/lib/contextual-action-panel/soho-contextual-action-panel.d.ts
@@ -32,6 +32,8 @@ interface SohoContextualActionPanelOptions {
   /** The string used as the title for the panel - not defaulted. */
   title?: string;
 
+  cssClass?: string;
+
   /** Settings to pass through to the modal */
   modalSettings?: SohoModalOptions;
 
--- a/src/app/contextual-action-panel/contextual-action-panel.demo.html
+++ b/src/app/contextual-action-panel/contextual-action-panel.demo.html
@@ -5,6 +5,9 @@
       <button soho-button (click)="openPanel()" data-modal="contextual-action-modal-1">Panel</button>
     </div>
     <div class="field">
+      <button soho-button (click)="openPanelCSS()" data-modal="contextual-action-modal-1">Panel CSS</button>
+    </div>
+    <div class="field">
       <button soho-button (click)="tabsVertical()"  data-modal="contextual-action-modal-1">CAP - Tabs Vertical</button>
     </div>
     <div class="field">
--- a/src/app/contextual-action-panel/contextual-action-panel.demo.ts
+++ b/src/app/contextual-action-panel/contextual-action-panel.demo.ts
@@ -111,6 +111,23 @@ export class ContextualActionPanelDemoComponent {
     }).open();
   }
 
+  openPanelCSS() {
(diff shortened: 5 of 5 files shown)


tmcconechy (approved):

ericangeles (approved):

whernebrink:
@tmcconechy, I pulled latest main, npm i (since ng 13 now), then npm run build, and then npm start. I can see the code and the exapmle, but when I inspect I do not see the cssClass. Did you when testing?

image

tmcconechy:
@tjamesallen15 @whernebrink - try again. i had to update the dev version. So you will need to:

- pull main
- npm i
- npm run start
- go to http://localhost:4200/ids-enterprise-ng-demo/contextual-action-panel
- open panel css and inspect

Also i think it has to be:

```
    this.panelRef = (this.panelService as any).contextualactionpanel(ContextualActionPanelComponent, this.placeholder)
      .modalSettings({ title: this.title, useFlexToolbar: true })
      .open()
      .cssClass('my-custom-panel-test')
 ```
 
 So you need cssClass after open because the open method will change the classes out in angular

**Comment by whernebrink** [comment]
@tmcconechy, Well, then what issue did this ticket fix? :) Because my initial issue #1215, having it after open() was the thing working, whereas I was saying it did ont work when adding it in the `modalOptions` (with the rest of the options like buttons), and nod with .cssClass() before the .open().

**Comment by tmcconechy** [comment]
hmm good question - @tjamesallen15 / @ericangeles ??

**Comment by tjamesallen15** [comment]
> hmm good question - @tjamesallen15 / @ericangeles ??

In my test class, we can see i called first the .cssClass() before .open()
image

**Comment by whernebrink** [comment]
@tjamesallen15, can you please verify, when running the demo example, that the `cssClass: 'my-custom-panel-test` is added to the `<div class="contextual-action-panel modal"> in the DOM`? Both for when added to `.cssClass('my-custom-panel-test)` before the `.open()` call, and when one has added it to the `.modalSettings({ cssClass: 'my-custom-panel-test' })`. Just want to make sure it is not just for me locally it is not added.

**Comment by whernebrink** [comment]
After a `npm prune && npm i` I can confirm that having `.cssClass()` before `.open()` works. However, it does not work when specified inside the modalSettings:

image

Classes added: 
image

**Comment by whernebrink** [comment]
@ericangeles Alright, I see. Yeah, that would be great to have a PR for since it is on the modal options interface for CAP already.

image
Perhaps there are other things too on the interface which is not yet on EP? Could be worth having a look at perpahs?

**Comment by ericangeles** [comment]
@whernebrink ,I would rather stick to the current `modalSettings` options listed in the CAP, but if you need anything that needs to expose in the `modalSettings` let us know.

**Comment by whernebrink** [comment]
@ericangeles, my screenshot is of the `SohoContextActionPanelRef`, which uses the SohoModalOpions.

**Comment by tmcconechy** [comment]
What if we made the modal work the same way so you need to apply the cssClass after opening. Personally i dont see why it would be a major problem as long as it works one way....

I guess the way its invoked is different and thats why. So is this really worth fixing further?

**Comment by whernebrink** [comment]
Having .cssClass() only after the .open() I don't think is the way to go, for a couple of reasons:
* It is not clear that this `.cssClass()` has to go after the .open()
* For standard modal dialogs .cssClass() is probably used before the .open(), and in the modalSettings, and I imagine this would be a breaking change for many teams.
* Using the dot notation with chainable functions e.g. `.buttons()`, `.id()` etc has already been deprecated, and the developers are told to use the modalSettings: 
image So it seems weird to rely on using that?

In my case, I don't mind usning .cssClass() before or after the open, or in the modalSettings (which I would prefer if it would have worked). But when you have four ways, I forgot to mention that you can actually pass in an 
`SohoContextualActionPanelOptions` as the third argument when you open the CAP dialog too):
```ts
this.panelRef = this.panelService.contextualactionpanel(
  ContextualActionPanelComponent,
  this.placeholder,
  { cssClass: 'my-custom-class-0' } // `SohoContextualActionPanelOptions`
)
      .modalSettings({ title: this.title, useFlexToolbar: true, cssClass: 'my-custom-class-1' }) // `SohoModalOptions`
      .cssClass('my-custom-class-2')
      .open()
      .cssClass('my-custom-class-3')
      .initializeContent(true);
```
it is hard for us developers to know; 1. Which approach is recommended and why (adding the @deprecated helps for e.g. `.buttons()`), 2. Why does it not work when the interface says it should?

So if I were to come with one suggestion, it would be to use the `modalSettings` which currently is of type `SohoModalOptions`, but omit the things that is not supported:

```ts
// Use SohoModalOptions as the base for the options but omit the properties we don't support:
type SohoContextualActionPanelModalOptions = Omit<SohoModalOptions, 'cssClass' | 'somethingMore'>;

/**
    /**
     * Sets the whole options block for this contextual action panel.
     *
     * @param modalSettings - the options to set.
     */
    modalSettings(modalSettings: SohoContextualActionPanelModalSettings): SohoContextualActionPanelRef<T>;
```
Same goes for the `SohoContextualActionPanelOptions`, which is what you can pass as the third argument to `this.panelService.contextualactionpanel(Component, placeholder, options)`, where e.g. `cssClass` is available but currently not working either. I'm not sure of the internals here, but if it is the same as the object we can pass to `modalSettings` that would be great, so that we can reuse the type with omitted properies. 
 
```ts
/**
     * Sets the whole options block for this contextual action panel.
     *
     * @param options - the options to set.
     */
    options(options: SohoContextualActionPanelOptions): SohoContextualActionPanelRef<T>;
```

Sorry for long reply, I do hope it is any good feedback which is helpful and could potentially increase the developer experience. 

Lastly. Thanks for your fantastic work! I really appreciaty you all for addressing many of these issues in a quick manner. I understand that spending time on these things is not the most important/prioritized thing, given we're moving towards the web components, but we are probably many teams that will be stuck to enterprise-ng for a good while. 😬

tmcconechy:
Ok @whernebrink so i guess then we don't need 4 ways to do this. So what if we just do it with modalSettings? Would that work for you here if fixed?

@ericangeles can we change this example to show setting it on modalSettings? If needed we could set the cssClass on some variable and feed it into settings on open if the timing isnt right?

tjamesallen15:
> Ok @whernebrink so i guess then we don't need 4 ways to do this. So what if we just do it with modalSettings? Would that work for you here if fixed?
> 
> @ericangeles can we change this example to show setting it on modalSettings? If needed we could set the cssClass on some variable and feed it into settings on open if the timing isnt right?

@tmcconechy  I've made another set of PRs for this one.
https://github.com/infor-design/enterprise/pull/6164
https://github.com/infor-design/enterprise-ng/pull/1250

PR #1320, opened 2022-06-08
Change HTML order of soho-button

LilBiggs:
Reorganized HTML structure of button to place the <span> for the button text in front of the <svg> tag. This is the order that the IDS jQuery library creates the button. Without this order, the notification badge is slightly off when setting to upper-right

**Related github/jira issue (required)**:
Fixes #1319 

**Steps necessary to review your pull request (required)**:
Before: 
image

After:
image

PR #1341, opened 2022-07-12
1340 - Add button alignment setting for icons, fix regressions

tmcconechy:

Fixes a regression that the icons "moved" to the wrong side. In doing this added a setting so this can be done voluntarily
and refixed the original issue: https://github.com/infor-design/enterprise-ng/pull/1320

Fixes #1340

- pull and run `npm run build:lib && npm run start`
- check out http://localhost:4200/ids-enterprise-ng-demo/button-badge (icons should be aligned still as per https://github.com/infor-design/enterprise-ng/pull/1320)
- check out http://localhost:4200/ids-enterprise-ng-demo/button (you can see the icons are now on the left again and there is a new example with them on the right via a setting)


ericangeles:
@tmcconechy do we need to have this setting `iconAlign` in EP?

Issue #1419, opened 2022-12-15
DataGrid | Pager - beforePaging and afterPaging not fired without a source on datagrid options


LucasLundJensen:
**Describe the bug**
The afterPaging and beforePaging events are only emitted from the datagrid if a source is attached in the datagridOptions.

**To Reproduce**
1. Create a datagrid without a source
2. Listen for the event and execute a console.log
Result: Nothing happens

Alternatively:
1. Remove the source from this repositories example
2. See console not emitting event when changing page

**Expected behavior**
The events should be fired regardless of a source being attached in the datagrids options or not.

**Version**
- ids-enterprise-ng: 14.7.1

**Platform**
 - OS Version: Windows 10
 - Browser Name Chrome
 - Browser Version  104.0.5112.102 64-bit

LucasLundJensen:
I'll create a pull request that exposes the events, just like it's done with the standalone pager component.

CindyMercadoReyes:
This issue is now resolved.

PR #1422, opened 2022-12-16
Add missing built-in pager events to datagrid

LucasLundJensen:
The built-in pager on Datagrid's dont properly expose the emitted events, the current "beforePaging" and "afterPaging" events are only fired when there's a source added to the datagrid.

This requests exposes all the standard events that are exposed like the standalone-pager.

Fixes https://github.com/infor-design/enterprise-ng/issues/1419

1. Build the project
2. Go to "Datagrid Paging Service" example
3. Open console
4. Use the different paging buttons and see events being fired in console.

Extra steps:
You can also check it works without the source being attached to the datagrid, disable it by uncommenting it.
Then click the "Set Pager to Page Two" and use the paging buttons (Kind of a hacky way, but it's the easiest with fewest changes)


ericangeles (approved):

tmcconechy (approved):
Looks good. Thanks @LucasLund

PR #1426, opened 2022-12-20
1418: Unified `soho-icons-empty` with ability to force new "colorful" icons

### LINKED ISSUE #1418: Ability to force new new empty state icons
We want to use the "new" new empty state icons, e.g `#icon-empty-new-project-new`. We'd like for *all* empty states in the app to use the new icons, without having to change every single `SohoEmptyMessageDirective.icon` input. This is especially important for micro-frontend apps such as Portal / Homepages, where we are not able to modify all of the source code (widgets, extensions).

Add an option *somewhere* that makes it so that e.g `soho-emptymessage icon="icon-empty-no-data"` is equivalent to `soho-emptymessage icon="icon-empty-no-data-new"`.

Some possible solutions:
1. Add a `SohoIconsEmptyNewNewComponent` that loads the the "new" new SVG:s with the "old" identifiers.
2. Add a  `[forceNewNewIcons]="true"` option to `SohoIconsEmptyNewComponent`, which does the same as above.

The only alternative (aside from implementing the above on our own) is to change all the `<soho-emptymessage icon="..."` in the code. This is not possible to do at exactly the same time for all external code (widgets) as it would involve similar changes to thousands of lines of code spread over many developers/teams/companies.

**Additional context**
I'm using the term "new new" somewhat tongue in cheek here, because I don't know what the "new" new icons should be called... 😉

anhallbe:

Adds ability to configure `soho-icons-empty` to use one of:
- `"classic"` (default for backwards compatibility) - Icons for classic themes
- `"new"` - Same as `soho-icons-empty-new`, so both the "old" new/uplift icons and the `-new` suffixed icons.
- `"colorful"` (Better name suggestions?) - Use only the `-new` suffixed icons.

This allows an app to use the "new" colorful icons with the `-new` suffix, but without having to specify the `-new` suffix. I.e the following:
```html
<soho-icons-empty icons="colorful"></soho-icons-empty>
<div soho-emptymessage icon="icon-empty-no-data">
```
is equivalent to this:
```html
<soho-icons-empty-new></soho-icons-empty-new>
<div soho-emptymessage icon="icon-empty-no-data-new">
```

Closes #1418 

- Check out the branch
- Run `npm run build:lib && npm start`
- Go to the "Empty Message" sample (and others that might display an empty-message icon).
- Switch themes, make sure it works.
- Use the "old" `<soho-icons-empty>` and `<soho-icons-empty-new>` without the new option to make sure it still works.

NOTE:
This PR is for now mostly for review/discussion. I have not added anything to the changelog or done any extensive testing yet.

NOTE 2:
When running the sample app locally, not all the icons are loaded initially sometimes. But as soon as the theme/color is switched it works. This *might* have been caused by me. But as far as I can tell the issue was there with the old components as well. Issue with angular 15 upgrade? WIll need investigation

**TODO**
- [x] Decide if "colorful" is a good name for the new icons
- [x] Decide what the default `icons` should be. Right now it is `classic` to avoid breaking changes.
- [x] Investigate why icons are not properly loaded when refreshing the demo app
- [ ] Currently there is no way to tree-shake the icons since they are only hidden by Angular during runtime. This means that e.g the "classic" icons template will always be included, even if they are never used.
- [x] Add changelog
- [x] Deprecate `soho-icons-empty-new`?


tmcconechy:
Should i be able to dynamically change them on the fly? I tried adding icons="colorful" to the dom and it didnt work.
Do you want to make the example show all the new ones?

- [ ]  Decide if "colorful" is a good name for the new icons
I'm fine with that - "new" is as you are saying overloaded
- [ ]  Decide what the default icons should be. Right now it is classic to avoid breaking changes.
Lean towards making the newest ones the default to encourage use
- [ ]  Investigate why icons are not properly loaded when refreshing the demo app
Not following - how do i see this?
- [ ]  Currently there is no way to tree-shake the icons since they are only hidden by Angular during runtime. This means that e.g the "classic" icons template will always be included, even if they are never used.
Can templates be loaded at runtime dynamically? https://blog.angular.io/angular-v15-is-now-available-df7be7f2f4c8#ecb8
- [ ]  Add changelog
Yes please
- [ ]  Deprecate soho-icons-empty-new
Yes just make a note in the change log to that effect.

tmcconechy:
@anhallbe did you want to finish this off or should we do it next time around?

anhallbe:
Hi @tmcconechy!
I've been away for the holidays. but now I'm back. I found the issue with the icons not being set correctly in the demo app (see: [cbafa06](https://github.com/infor-design/enterprise-ng/pull/1426/commits/cbafa0617b236e495b7d106b17a7e393c413125c))

I'll see if I can get lazy-loading/tree-shaking to work, but other than that I'd say this PR is pretty much ready.

Regarding defaults: Should we make `"colorful"` the default instead of `"classic"`? It would be a breaking change, since apps that currently use `soho-icons-empty` will get the colorful ones. But on the other hand it's easy to go back to "classic".

tmcconechy:
Yes i dont mind if we make the new ones "colorful" the default. This was the design direction we got. I made all the examples try to show the new ones. lets just add a not in change log however.

Sounds good :)

anhallbe:
I made the "colorful" icons the default and updated jsdoc & changelog.

tmcconechy (approved):
Looks good to me @anhallbe thanks!
cc @ericangeles

ericangeles (approved):
Looks good! Thanks @anhallbe 👍🏼

PR #1503, opened 2023-06-13
Update soho-monthview.component.ts, exposed showMonth() method

yousef-elkaisi:
Added public showMonth() method to allow the month to be manually set.

This allows us to manually set the month and year of the monthview component.

Create a monthview component and call the showMonth() method through the viewchild accessor. Confirm that the month is changed to the value that was set.

### DIFF
--- a/projects/ids-enterprise-ng/src/lib/monthview/soho-monthview.component.ts
+++ b/projects/ids-enterprise-ng/src/lib/monthview/soho-monthview.component.ts
@@ -645,8 +645,13 @@ export class SohoMonthViewComponent implements AfterViewChecked, AfterViewInit, 
     });
   }
   
+  /**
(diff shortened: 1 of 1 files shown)


tmcconechy (approved):

tmcconechy (changes requested):
@yousef-elkaisi Seems to not compile https://github.com/infor-design/enterprise-ng/actions/runs/5257447799/jobs/9501402032?pr=1503#step:5:32

ericangeles:
@yousef-elkaisi, you might want to consider using optional chaining. For example, you could use `this.monthview?.showMonth(month, year);`.

yousef-elkaisi:
@tmcconechy @ericangeles I added a null check and casted monthview as 'any'. This should fix the compile issues.

PR #1548, opened 2023-08-22
7822 - Added listcontextmenu event

tmcconechy:

Adds a listcontextmenu event which is used to add data for right click, open in a new window functionality for Landmark.

Fixes https://github.com/infor-design/enterprise/issues/7822

- build branch https://github.com/infor-design/enterprise/pull/7847 and copy it to the node_modules folder
- open http://localhost:4200/ids-enterprise-ng-demo/dropdown
- open the module switcher menu and right click any element (for example Landing Page Designer)
- then open in new tab -> check the url it should be injected with the element name


tmcconechy:
cc @krishollenbeck @EdwardCoyle

PR #1576, opened 2023-10-25
7995 - Fix on dataset creating a new array on update

yohannahbautista:

Checked on IDS and dataset is in sync with model data when reproducing the scenario. It's in NG where it creates a copy of the data array instead of using pass by reference when trying to change the dataset, using updateDataset instead of updated (which is called when gridOptions is changed) keeps the reference instead of creating a new array.

Closes https://github.com/infor-design/enterprise/issues/7995

- Pull this branch, build and run the app
- Go to: http://localhost:4200/ids-enterprise-ng-demo/datagrid-update-dataset
- Add rows, should update. In console, should print success
- Change dataset with dropdown
- Add rows, should update. Console should print success
- Change back to previous dataset, should be updated (added rows should still be present)


ericangeles (approved):

tjamesallen15 (approved):

tmcconechy:
@yohannahbautista looks like some conflicts.

PR #1713, opened 2024-06-04
1702 - Fix readonly to work with disabled property in Datepicker

### LINKED ISSUE #1702: Datepicker: readonly flag doesn't work work from template
**Describe the bug**
Setting read only on a date picker component from the template on load is broken.

**To Reproduce**

1. Open the demo app
2. Go to the datepicker demo
3. See that the read only field is editable

**Expected behavior**
Read only should be able to be set for load with the [readonly] input.
readonly="true"
[readonly]="readonlyFlag"

**Version**
ids-enterprise-ng: latest

**Screenshots**
image

**Platform**
All

ericangeles:

Readonly and disabled state are associated together when building the control (for disabled state). Which means that the `this.readonly` and `this.disabled` will be set to true if the control is disabled. We need to get the parent element to check if the control is disabled `.is-disabled` (Coming from the jQuery library). This is connected to `setDisabledState()` function.

When it runs to `setDisabledState()`, it went back to normal state. This PR will fix that.

Closes #1702 

- Go to http://localhost:4200/ids-enterprise-ng-demo/datepicker
- Check the `Read Only` element, datepicker should be in readonly state
- Check also the` Disable` element, datepicker should be in disabled state


tmcconechy (approved):

tmcconechy (changes requested):
@ericangeles can you check those tests. Seems like disable with the form group isnt working maybe?

